PR #47, opened 2020-11-25
Set quality/directional_shadow/size=16384

comick:
Lowering to default value of 4094 to prevent GL_OUT_OF_MEMORY in glTexImage2D

This happens on my radeon card (r600 on linux).
Having a godot default here improves compatibility.

As a moving forward activity, more renderer things can be made configurable. (eg: FSAA 4x by default is a bit heavy)
Anything against receiving pull requests in that direction?

### DIFF
--- a/src/project.godot
+++ b/src/project.godot
@@ -152,6 +152,5 @@ nextTable={
 [rendering]
 
 threads/thread_model=2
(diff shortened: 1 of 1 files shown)


Jean28518:
Thank you for contributing!

PR #48, opened 2020-11-26
Consistent camera movements with mouse and wasd keys at any fps

comick:
refDelta is repeated and the camera movement itself is.
I preferred to avoid refactoring of things since the change is minimal for now.


comick:
This change basically brings the same free cameras handling experience at any fps

PR #57, opened 2020-12-08
Player cameras

comick:
factored FreeCamera.gd into configurable generale purpose camera script Camera.gd. Later I think could be used for both cabin and outer views. Wanted to keep change thin and focused for this first step.
Player cameras are activated with keys "3" to "8" ("9" I'd like to keep free for panoramic mode where train is seen from stations and arbitrary scenario points).
An example player camera is placed in JFR1_Red first wagon, right side, child height.
It has acceleration just like cabin camera. Adding more camera nodes to the group "PlayerCameras" is how more are added.


Jean28518:
Very nice, thank you!

PR #70, opened 2021-01-05
Add police and test models with animations

bodamat:
For now, I imported models police and test model with animations. The test model has not fixed animations. Police have fixed animations, but I found new bugs and fix them later. Police don't have upstairs and downstairs animations, because mixamo doesn't have a good one. These animations need fix and I also do it later.


Jean28518:
Thanks! I am reviewing it..

Jean28518:
They are looking good!
But under which license are the animations and the object?

bodamat:
> They are looking good!
> But under which license are the animations and the object?

I don't know which licence. I only know it's free

https://helpx.adobe.com/ua/creative-cloud/faq/mixamo-faq.html

Jean28518:
That's the problem, I also just found this FAQ... 
So maybe it would be better, if we create these Animations at ourselves, maybe in blender.
What do you think about it?
Did you check out https://gamefromscratch.com/makehuman-1-2-released/
That could be more suitable to us, I guess

bodamat:
> That's the problem, I also just found this FAQ...
> So maybe it would be better, if we create these Animations at ourselves, maybe in blender.
> What do you think about it?
> Did you check out https://gamefromscratch.com/makehuman-1-2-released/
> That could be more suitable to us, I guess

Sorry for the long reply! I haven't seen this video before. I think It's a good idea because it's open source. I like it 😍 But it's may more difficult. Because I'm a programmer, not an artist. I don't know that I can create good animations and it may take a lot of time. 
For a quick result mixamo good for now. For a long result, we can create our own animations. For both results, we can create a character in this program and animations from mixamo. What do you think? Do we have enough time?

I like "MakeHuman" for creating children 😅

Jean28518:
Oh yeah, we have that time..
The thing is, I don't know how to license these files from mixamo correctly.. 
Before we run in licensing issues, I think let's do that ourselves. Then maybe we learn some new things ;)
If you you don't want to do the animation, just let me know it, and I will do it in the future..
If you read the new devlog we have enough time for the next release ;) https://www.libre-trainsim.de/news

bodamat:
Really? This is a so huge time we have. Maybe we can do it more quickly? 😉 Go to the discord to discuss more detail and plan?

Yeah, you're right about Mixamo. We can create our own animation if we have a lot of time. Maybe it's even better. 
I want to create, but I don't know if my knowledge is good for that, and I don't know I'll have enough time. Discuss in the Discord 😉

Jean28518:
`Maybe we can do it more quickly? wink `
Maybe ;)) But I am planning an extension for Wolfsfurt Buchenhain, and a new Highspeed track.. That could take a lot of time ;))

bodamat:
First I need to learn Godot because I don't know it well. Then I can help more :)

bodamat:
I think I can close this pull request, because you did much better :)

PR #88, opened 2021-08-19
Fixed small grammar mistake in Ingame Translations file

akuspel:
Changed "later" to "late" in src/Translations/Ingame to fix small grammar error pointed out in Issues


Jean28518:
Thanks for contributing!

Issue #77, opened 2021-02-25
Possibility of driving backwards


simfann:
Sometimes it's very helpful to have a possibility to drive backwards. For example when you have missed a station or you have braked too late at the station or at a red signal and want to correct that situation by driving back. Else the scenario will come to an end without a possibility to cancel your mistake.

Jean28518:
Thanks for your idea!

I will think about it, how to implement this :)

DasCapschen:
I will look into this and start working on it :)

PR #95, opened 2021-08-21
Driving Backwards

DasCapschen:
Implement driving backwards.  Solves #77.  
There are quite a lot of changes to core driving logic, so you probably want to review this properly.  
So far I have done some limited testing on the tutorial map, but it seemed to work fine there.

## To Do:
- [x] Change station behaviour if train missed a station
- [x] Implement new feature to autopilot (has to "press" F once)
- [x] Enabling Debug Mode: Should set train to forward
- [x] Hint if you throttle up and the reverser is neutral
- [x] Controller mappings
- [x] Mobile version needs buttons for the reverser

## Known issues:   
- [x] Going across a green light at the end of a station turns it red, going backwards over it again does *not* turn it green again. This probably requires resetting "nextStation" or so. I haven't yet looked into station code. Will do soon.
- [x] Signals are handled before the train drives over sometimes (tested on Hainfurt -> Buchenhain). Seems to be in connectionwith the start distance on a rail in backward direction at beginning of a scenario
- [x] The go back message does not include the controls


Jean28518:
Thanks, I will review it!
That issue with the signal: I wouldn't reset any station variable because signals are not connected to stations in any way.

I would implement in this way: The train stores every overdriven signal in an array. If the train detects a signal in front, which is stored in this array, the train should turn that signal status to 1.

DasCapschen:
Okay, here is a short summary on how I implemented it:
- I store all signal names and route positions in arrays when baking the route. These arrays are already sorted, so `baked_route_signalNames[0]` will be the first signal the player passes.
- I keep track of the upcoming signal with an index `nextSignalIndex`
- in `check_signals` I check the player position on route against the signal position on route, and if passed, `nextSignalIndex += 1` (if driving backwards `nextSignalIndex -= 1`)
- I then find the last signal (with index < nextSignalIndex) of type "Signal" and call `giveSignalFree()`

Looks like `giveSignalFree()` is the problem, it only works if the signal is a block signal, but the signal at the end of the station is not. I guess I should just use `status = 1` right away.

Oh and I didn't yet check if you can drive past a station, then reverse into it and still service it. I think the game tells you "you drove past a station, go to the next one". So we would probably need to reset the station when reversing into it.

Jean28518:
Very nice and clean code!

At newer contributions I try to respect the following code guidelines: https://www.gdquest.com/docs/guidelines/best-practices/godot-gdscript/
Would be great if you could do this too in the future :-) 

Could you please explain me the difference between `distance` and `distanceOnRoute`?

DasCapschen:
Oh, of course. I can do a bit of refactoring to follow the code guidelines, no problem. :)

Oh, `distanceOnRoute` is `distance`, I actually just renamed the variable, because there was already `distanceOnRail` and I got confused by what `distance` is, so I thought calling it `distanceOnRoute` makes it a bit more clear.

Jean28518:
Okay, yeah, good Idea!

Jean28518 (commented):
Code is okay in general :+1:

DasCapschen:
I tested it on Hainfurt now and also had the AI and Autopilot problems, that is fixed now :)
Station signals (`blockSignal = false`) will now also turn green.

I will look into signals being handled too early now :+1:

DasCapschen:
Signals should no longer trigger early. This was indeed due to wrong initial distanceOnRoute, if `forward = false` for startRail.

Jean28518:
Why did you made the reverse state global? :)

DasCapschen:
Because it is accessed from multiple files, not just in Player.gd, and it's nicer to just write `ReverserState.FORWARD` than `player.ReverserState.FORWARD`, because really it's a constant, and not something intrinsic to the player.   
Remember this is just the Enum for the possible values, the current value of the player is `player.reverser` and not global :)

DasCapschen:
I have now refactored the Station logic, which makes it easier to read and understand, and I've also changed it a little.

The station now only counts as missed if the *whole* train went past the station. (`distance-distanceOnStationBeginning > stationLength` changed to `distance-distanceOnStationBeginning > stationLength + length`)

Just like when the end of train is not yet in the station, when the train drives past the station a bit, `whole_train_in_station` is false, and the player is told to reverse into the station and try again.

Also fixed a bug where people got out of the train when the player passed the station signal and opened the doors, but the train was not yet fully in the station. (people would walk in mid air)

src/Resources/Basic/Screens/ScreenReverser/Reverser.gd
```diff
@@ -0,0 +1,6 @@
+extends Control
+
+func _ready():
+	$"VBoxContainer/Forward".set_text(TranslationServer.translate("FORWARD"))
+	$"VBoxContainer/Neutral".set_text(TranslationServer.translate("NEUTRAL"))
+	$"VBoxContainer/Reverse".set_text(TranslationServer.translate("REVERSE"))
```
HaSa1002:
```suggestion
	$VBoxContainer/Forward.set_text(tr("FORWARD"))
	$VBoxContainer/Neutral.set_text(tr("NEUTRAL"))
	$VBoxContainer/Reverse.set_text(tr("REVERSE"))
```

src/addons/Libre_Train_Sim_Editor/Data/Modules/TrainInfo.tscn
```diff
@@ -239,11 +239,36 @@ __meta__ = {
 }
 actions = [ "doorClose", "doorRight", "doorLeft" ]
 
-[node name="Acceleration-0" type="HBoxContainer" parent="ScrollContainer/VBoxContainer"]
+[node name="Reverser" type="HBoxContainer" parent="ScrollContainer/VBoxContainer"]
 margin_top = 287.0
 margin_right = 753.0
 margin_bottom = 317.0
 
+[node name="dot" type="TextureRect" parent="ScrollContainer/VBoxContainer/Reverser"]
```
HaSa1002:
```suggestion
[node name="Dot" type="TextureRect" parent="ScrollContainer/VBoxContainer/Reverser"]
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Enums/ReverserState.gd
```diff
@@ -0,0 +1,7 @@
+class_name ReverserState
+
+enum {
+    FORWARD = 1,
+    NEUTRAL = 0,
+    REVERSE = -1
```
HaSa1002:
```suggestion
    FORWARD = 1,
    NEUTRAL = 0,
    BACKWARD = -1
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -471,9 +482,11 @@ func getSpeed(delta):
 		return
 	var lastspeed = speed
 	## Slope:
-	var currentSlope = currentRail.get_heightRot(distanceOnRail)
+	var currentSlope = currentRail.get_heightRot(distance_on_rail)
 	if not forward:
-		currentSlope = - currentSlope
+		currentSlope = -currentSlope
+	if reverser == ReverserState.REVERSE:
```
HaSa1002, line 499:
```suggestion
	if reverser == ReverserState.BACKWARD:
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -501,67 +514,62 @@ func getSpeed(delta):
...
-			change_to_next_rail()
-	else:
-		drivenDistance = speed * delta
-		distanceOnRail -= drivenDistance
-		distance += drivenDistance
-		if distanceOnRail < 0:
-#			drivenDistance = 0 - distanceOnRail
-			change_to_next_rail()
+	var driven_distance = speed * delta
+	if reverser == ReverserState.REVERSE:
```
HaSa1002, line 532:
```suggestion
	if reverser == ReverserState.BACKWARD:
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -501,67 +514,62 @@ func getSpeed(delta):
...
-				signals.erase(signalname)
-			if not forward and signalname != "":
-				handle_signal(signalname)
-				signals.erase(signalname)
 	
-	if forward:
-		distanceOnRail -= currentRail.length
+	if forward and (reverser == ReverserState.FORWARD):
+		distance_on_rail -= currentRail.length
+	if not forward and (reverser == ReverserState.REVERSE):
```
HaSa1002, line 557:
```suggestion
	if not forward and (reverser == ReverserState.BACKWARD):
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -501,67 +514,62 @@ func getSpeed(delta):
...
+		distance_on_rail -= currentRail.length
+
 	if not ai:
 		print("Player: Changing Rail...")
-	routeIndex += 1
-	if baked_route.size() == routeIndex:
+
+	if reverser == ReverserState.REVERSE:
```
HaSa1002, line 563:
```suggestion
	if reverser == ReverserState.BACKWARD:
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -501,67 +514,62 @@ func getSpeed(delta):
...
+
 	var new_radius = currentRail.radius
 	if forward:
 		new_radius = -new_radius
 	
-	if not forward:
-		distanceOnRail += currentRail.length
+	if not forward and (reverser == ReverserState.FORWARD):
+		distance_on_rail += currentRail.length
+	if forward and (reverser == ReverserState.REVERSE):
```
HaSa1002, line 582:
```suggestion
	if forward and (reverser == ReverserState.BACKWARD):
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -676,165 +684,231 @@ func handleCamera(delta):
...
-	for signalname in signals.keys():
-		if forward and signalname != "" and signals[signalname] < distanceOnRail:
-			handle_signal(signalname)
-			signals.erase(signalname)
-		if not forward and signalname != "" and signals[signalname] > distanceOnRail:
-			handle_signal(signalname)
-			signals.erase(signalname)
-			
-func handle_signal(signalname):
+	if reverser == ReverserState.REVERSE:
```
HaSa1002, line 701:
```suggestion
	if reverser == ReverserState.BACKWARD:
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -676,165 +684,231 @@ func handleCamera(delta):
...
 				wagonI.sendPersonsToDoor(platformSide, stations["leavingPersons"][current_station_index]/100.0)
-	elif signal.type == "Speed":
-		currentSpeedLimit = signal.speed
-	elif signal.type == "WarnSpeed":
-		print(name + ": Next Speed Limit: "+String(signal.warn_speed))
-	elif signal.type == "ContactPoint":
-		signal.activateContactPoint(name)
+		
+	elif signal_passed.type == "Speed":
+		if reverser == ReverserState.REVERSE:
```
HaSa1002, line 793:
```suggestion
		if reverser == ReverserState.BACKWARD:
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -676,165 +684,231 @@ func handleCamera(delta):
...
 
 
 
 ## For Station:
-var GOODWILL_DISTANCE = 10 # The distance the player can overdrive a station, or it's train end isn't in the station.
-var endStation = false
-var stationBeginning = true
+var GOODWILL_DISTANCE = 10 # distance the player can overdrive a station, or it's train end isn't in the station.
+var is_last_station = false # if this is the last station on the route
+var is_first_station = true # if this is the first station (where the player spawns)
```
HaSa1002, line 809:
```suggestion
var is_last_station := false # if this is the last station on the route
var is_first_station := true # if this is the first station (where the player spawns)
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -1221,14 +1307,20 @@ var autoPilotInStation = true
 var updateNextSignalTimer = 0
 func updateNextSignal(delta):
 	if nextSignal == null:
-		if get_all_upcoming_signalPoints_of_types(["Signal"]).size() == 0: return
-		nextSignal = world.get_node("Signals").get_node(get_all_upcoming_signalPoints_of_types(["Signal"])[0])
+		var upcoming = get_next_Signal()
+		if upcoming == null: return
```
HaSa1002:
```suggestion
		var upcoming = get_next_signal()
		if upcoming == null:
			return
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -1221,14 +1307,20 @@ var autoPilotInStation = true
...
+		nextSignal = world.get_node("Signals/"+upcoming)
 		updateNextSignalTimer = 1 ## Force Update Signal
 	updateNextSignalTimer += delta
 	if updateNextSignalTimer > 0.2:
 		distanceToNextSignal = get_distance_to_signal(nextSignal.name)
 		updateNextSignalTimer = 0
 
+func get_next_Signal():
```
HaSa1002:
```suggestion
func get_next_signal():
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -1254,14 +1346,18 @@ var nextStationNode = null
...
-	distanceToNextStation -= speed*delta
 	if nextStationNode == null:
-		if get_all_upcoming_signalPoints_of_types(["Station"]).size() > 0:
-			nextStationNode = world.get_node("Signals").get_node(get_all_upcoming_signalPoints_of_types(["Station"])[0])
-			nextStationNode.set_waiting_persons(stations["waitingPersons"][0]/100.0 * world.default_persons_at_station)
-			distanceToNextStation = get_distance_to_signal(nextStationNode.name) + nextStationNode.stationLength
-
-
+		var upcoming = get_next_Station()
+		if upcoming == null: return
```
HaSa1002:
```suggestion
		var upcoming = get_next_station()
		if upcoming == null:
			return
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -1254,14 +1346,18 @@ var nextStationNode = null
...
+		nextStationNode = world.get_node("Signals").get_node(upcoming)
+		nextStationNode.set_waiting_persons(stations["waitingPersons"][0]/100.0 * world.default_persons_at_station)
+	distanceToNextStation = get_distance_to_signal(nextStationNode.name) + nextStationNode.stationLength
+
+func get_next_Station():
```
HaSa1002:
```suggestion
func get_next_station():
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -1307,7 +1403,7 @@ func autopilot(delta):
 
 			
 	## Open Doors:
-	if (currentStationName != "" and speed == 0 and not isInStation and distance-distanceOnStationBeginning>=length):
+	if (currentStationName != "" and speed == 0 and not isInStation and distance_on_route-distanceOnStationBeginning>=length):
```
HaSa1002:
```suggestion
	if (currentStationName != "" and speed == 0 and not isInStation and distance_on_route - distanceOnStationBeginning >= length):
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -1395,28 +1491,27 @@ func toggle_front_light():
...
-var freeLastSignalBoolean = false
-var newSignalDistance = 0
-func freeLastSignalAfterDrivenTrainLength(): # Called, when overdrove the next signal
-	lastDrivenSignalTmp = lastDrivenSignal
-	newSignalDistance = distance
-	freeLastSignalBoolean = true
+var signal_to_free = null
+var signal_to_free_distance = 0
+func freeSignalAfterDrivenTrainLength(signal_instance): # Called, when overdrove the next signal
+	signal_to_free = signal_instance
```
HaSa1002, line 1519: I suspect we may leak memory here, so just in case:
```suggestion
	assert(signal_to_free == null)
	signal_to_free = signal_instance
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -1504,18 +1599,37 @@ func check_overdriving_a_switch():
...
+
+	match reverser:
+		ReverserState.FORWARD:
+			if change < 0:
+				reverser = ReverserState.NEUTRAL
+				jAudioManager.play_game_sound("res://Resources/Basic/Sounds/click.ogg")
+		ReverserState.NEUTRAL:
+			jAudioManager.play_game_sound("res://Resources/Basic/Sounds/click.ogg")
+			if change > 0: reverser = ReverserState.FORWARD
+			else: reverser = ReverserState.REVERSE
```
HaSa1002, line 1648:
```suggestion
		ReverserState.NEUTRAL:
			jAudioManager.play_game_sound("res://Resources/Basic/Sounds/click.ogg")
			if change > 0:
				reverser = ReverserState.FORWARD
			else: 
				reverser = ReverserState.BACKWARD
```
or inline:
```suggestion
		ReverserState.NEUTRAL:
			jAudioManager.play_game_sound("res://Resources/Basic/Sounds/click.ogg")
			reverser = ReverserState.FORWARD if change > 0 else ReverserState.BACKWARD
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -1504,18 +1599,37 @@ func check_overdriving_a_switch():
...
+		ReverserState.REVERSE:
+			if change > 0:
+				reverser = ReverserState.NEUTRAL
```
HaSa1002, line 1651:
```suggestion
		ReverserState.BACKWARD:
			if change > 0:
				reverser = ReverserState.NEUTRAL
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Rail.gd
```diff
@@ -69,7 +69,7 @@ onready var world = find_parent("World")
 onready var buildings = world.get_node("Buildings")
 
 
-var attachedSignals = {}
+var attachedSignals = []
```
HaSa1002, line 72: Why not a dictionary?

src/addons/Libre_Train_Sim_Editor/Data/Scripts/TrainInfo.gd
```diff
@@ -65,7 +65,14 @@ func update_info(player):
 		else:
 			$"ScrollContainer/VBoxContainer/Brakes-0/dot".texture = green
 			$"ScrollContainer/VBoxContainer/Brakes-1/dot".texture = green
-	
+
+	## Reverser:
+	if player.reverser == ReverserState.NEUTRAL:
+		$"ScrollContainer/VBoxContainer/Reverser/dot".texture = red
+	else:
+		$"ScrollContainer/VBoxContainer/Reverser/dot".texture = green
```
HaSa1002:
```suggestion
		$ScrollContainer/VBoxContainer/Reverser/Dot.texture = red
	else:
		$ScrollContainer/VBoxContainer/Reverser/Dot.texture = green
```
Why do we set the texture and not just use `self_modulate` or `modulate`?

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Wagon.gd
```diff
@@ -107,43 +107,60 @@ func _process(delta):
...
-			if distanceOnRail > currentRail.length:
-				change_to_next_rail()
-		else:
-			drivenDistance = speed * delta
-			distanceOnRail -= drivenDistance
-			distance += drivenDistance
-			if distanceOnRail < 0:
-				change_to_next_rail()
+		var driven_distance = speed * delta
+		if player.reverser == ReverserState.REVERSE:
```
HaSa1002, line 124:
```suggestion
		if player.reverser == ReverserState.BACKWARD:
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Wagon.gd
```diff
@@ -107,43 +107,60 @@ func _process(delta):
...
+#       can we have a single method that both of them use?
 func change_to_next_rail():
-	if forward:
-		distanceOnRail -= currentRail.length
-	routeIndex += 1
-	currentRail =  world.get_node("Rails").get_node(baked_route[routeIndex])
-	forward = baked_route_direction[routeIndex]
+	if forward and (player.reverser == ReverserState.FORWARD):
+		distance_on_rail -= currentRail.length
+	if not forward and (player.reverser == ReverserState.REVERSE):
```
HaSa1002, line 141:
```suggestion
	if not forward and (player.reverser == ReverserState.BACKWARD):
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Wagon.gd
```diff
@@ -107,43 +107,60 @@ func _process(delta):
...
+		distance_on_rail -= currentRail.length
+
+	if player.reverser == ReverserState.REVERSE:
```
HaSa1002, line 144:
```suggestion
	if player.reverser == ReverserState.BACKWARD:
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Wagon.gd
```diff
@@ -107,43 +107,60 @@ func _process(delta):
...
+	currentRail =  world.get_node("Rails").get_node(baked_route[route_index])
+	forward = baked_route_direction[route_index]
+
 	updateSwitchOnNextChange()
-	if not forward:
-		distanceOnRail += currentRail.length
+
+	if not forward and (player.reverser == ReverserState.FORWARD):
+		distance_on_rail += currentRail.length
+	if forward and (player.reverser == ReverserState.REVERSE):
```
HaSa1002, line 161:
```suggestion
	if forward and (player.reverser == ReverserState.BACKWARD):
```

HaSa1002 (changes requested):
I only did brain compilation but looks good overall. Just some style stuff

src/addons/Libre_Train_Sim_Editor/Data/Scripts/TrainInfo.gd
DasCapschen: Because the texture is not white. iirc, it has a white outline with the color in the middle. We cannot modulate that.

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
DasCapschen: I don't think this assertion is necessary. We are not freeing the signal as in deleting it, we are only setting it to "green".
So there is no memory which is allocated or can be leaked. We are only remembering a pointer to a node basically, and then overriding it.

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Rail.gd
DasCapschen: Because I want to sort them, so they need to be in some definite order. A dictionary is not ordered in any way, and cannot be accessed over indices, only over keys. This is not useful in this case (check where attachedSignals is used in Player.gd)

DasCapschen:
@HaSa1002 
I've merged some of your changes, but why all the changes of "Reverse" to "Backward"?

HaSa1002:
I believe, you rather drive backwards than driving reverse

DasCapschen:
Well, in a car the "Rückwärtsgang" is the "Reverse gear", so I'd say you are driving "in reverse", rather than "backwards".
It doesn't really make much of a difference, I think.

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
HaSa1002: oh my bad. I am still desperate to find these memory leaks ^^

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Rail.gd
HaSa1002: good point

HaSa1002:
Then it is personal preference, I guess. I have no strong opinion about that as I am not a native speaker. Both is understandable IMO. Func fact: There is a comment in your code under a `.REVERSE` talking about backwards driving.

HaSa1002:
I forgot to ask you to rename your files to `snake_case` per https://docs.godotengine.org/en/stable/getting_started/workflow/project_setup/project_organization.html#style-guide

DasCapschen:
Will do :+1:

HaSa1002:
* The go back message does not include the controls
* No hint is shown if you throttle up and the reverser is neutral
* In the middle on control board the directions are all displayed, but don't appear to have a function
* `R` and `F` feel inverted, because R is above F but reverses
* Controller mappings are missing (maybe, use DPAD, but we run out of buttons on controllers, so we need to find a better solution anyway)

HaSa1002:
The mobile version needs buttons for the reverser as well

DasCapschen:
>     * The go back message does not include the controls
>     * No hint is shown if you throttle up and the reverser is neutral
Will fix these, no problem :)


>     * Controller mappings are missing (maybe, use DPAD, but we run out of buttons on controllers, so we need to find a better solution anyway)
There will be #108 at some point, which might help reduce controller mappings.
I will check what bindings TSW2 and TS2021 use.


>     * The mobile version needs buttons for the reverser as well
This will need UI work. Don't want to overload a tiny mobile screen with more and more controls.
I think eventually the mobile will *need* to be a subset of the PC version (like forcing simple mode)


>     * In the middle on control board the directions are all displayed, but don't appear to have a function
There are the 3 possible directions and a knob next to it (might be hard to see, the cabin lighting is awfully dark) that points to the selected direction. This will be useful for #108.


>     * `R` and `F` feel inverted, because R is above F but reverses
This is a but of a weird one. Yes, R being above F, you'd think that *R* goes forward and *F* goes backwards, but also **R**everse and **F**orward. So I'm not sure which is less confusing 😅

HaSa1002:
> This will need UI work. Don't want to overload a tiny mobile screen with more and more controls.
I think eventually the mobile will need to be a subset of the PC version (like forcing simple mode)

I think #108 will fix that, but not the controller stuff, as you don't want to use a cursor through a controller. I am thinking about a radial or quick slot menu.

DasCapschen:
I think #108 will work fine on controllers. I was thinking of doing it like TSW2 (they have an xbox version on which this works fine):

you don't enable the mouse cursor and move that around, rather you have a little "crosshair" in the center of your screen and whatever control is below that gets selected. When you press the interact button, the view gets locked and moving the view stick / mouse around then moves the control.

On a mobile, of course you could do all that with your finger.

I am not opposed at all to a radial menu though, that is actually a really good idea! Even for PC. 
Although it should go into it's own issue / PR :)

DasCapschen:
About the mobile HUD, I've just had an idea:
We have the up and down button, which are used to control the combined brake / acceleration lever. 
What if we also make them control the reverser? Like this:
- player is stopped (speed == 0)
- player clicks arrow up -> Reverser changes to FORWARD -> click / hold arrow up more -> accelerate
- player clicks arrow down -> Reverser changes to REVERSE -> click / hold arrow down more -> accelerate

And then while driving: 
- if FORWARD: arrow down -> brake
- if REVERSE: arrow up -> brake

Basically making the arrows choose the direction you want to go in.

DasCapschen:
I would say this merge request is done. 
Feel free to test it extensively before you merge it. :)

HaSa1002 (approved):
You can't close the doors via controller anymore, but not a blocker since we need to rework controller input anyhow.
LGTM.

DasCapschen:
You should still be able to close the doors. I changed to behaviour so that when you press "doors right" and the doors are already open, they will then close.
But I thought about reworking how the doors work anyways :smile:

HaSa1002:
A good to know. But yeah not a problem really, as it was added in this dev version

PR #100, opened 2021-08-24
Remove duplicated stoneplate import file

HaSa1002:
This file caused import failure on Windows machines because their filesystem is case insensitive


Jean28518:
Thank you very much!

HaSa1002:
oh nice. I can finally update my branch without having to do hard resets all the time xD

PR #102, opened 2021-08-25
Fix error messages produced by the engine

DasCapschen:
This fixes the error messages the engine sends while playing, like `set_owner: owner != valid`.  
These changes should not have any noticable effect, other than removing error messages.


Jean28518:
Thank you very much!
Finally this != owner message is gone :)

PR #110, opened 2021-08-28
Add Controller Input

### LINKED ISSUE #105: Add controller support
This should be more or less simple using actions. Maybe we want to add controller specific controls to benefit from the axis of controllers.

*I will take a look at this this evening*

HaSa1002:
I also did some refactoring. For example, you can shorten `TranslationServer.translate()`to `tr()`, use `_unhandled_input_event` instead of `_process` to process input actions, and use GDScript format strings, to build strings.

THe change itself:
I added bindings for controllers (the easy part) and a new class called InputLabel that changes the input based on the device and supplied action. In order to insert any control into the translations they need to use format strings as well. Luckily `%s` converts anything that can be casted to a string (which is everything).

The detection of the used controller is incomplete, so everytime somebody reports a new GUID, it has to be added or we go and grab a SDL controller DB dumb an integrate that or we add a Godot Method with the most common gamepad types.

Lastly we should implement input settings. It should be relativly easy now, but I wanted to finish this PR to get into bed and back to my primary work. Please review carefully, as I had to a rebase way to late in the night.

Fixes #105


Jean28518:
Thanks for your pull request!
But you messed up the controls over keyboard..
Sorry, please fix that.

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Player.gd
```diff
@@ -402,17 +401,17 @@ func _input(event):
...
 			soll_command -= 0.7 * delta
 		if soll_command >= 1:
 			soll_command = 1
 		if soll_command <= -1:
 			soll_command = -1
-		if Input.is_action_pressed("ui_left"):
+		if Input.is_action_pressed("brake-"):
 			soll_command = 0
-		if Input.is_action_pressed("ui_right"):
+		if Input.is_action_pressed("brake+"):
```
HaSa1002, line 414: Do you mean that one?
The behaviour is the same

Jean28518:
Just drive the tutorial track on this branch with Easy Mode and arrow keys...
That doesn't work anymore.

WASD Steering is also different to the version of master branch...

HaSa1002:
I fixed the arrow key steering. Maybe I am blind, but how is WASD steering different? Are the acceleration or break axis inversed?

HaSa1002:
Your last commit removed WASD controls completly. Was that planned?

Jean28518:
Yes, that was planned :)

PR #127, opened 2021-09-03
Refactor platformSide

DasCapschen:
- platformSide is now an Enum (NONE, LEFT, RIGHT, BOTH)
- `platformSide` was renamed to `platform_side` to match coding guidelines
    + this requires renames in Scenes as well, or setting the attribute again


HaSa1002 (approved):
lgtm

PR #136, opened 2021-09-16
fix: Signal VisualInstance wrong initial state

DasCapschen:
The VisualInstance will show the wrong state, when a non-default state is set for the signal at spawn time. 

Eg. Signal status is set to green in the Scene file, but the VisualInstance will be red.


HaSa1002 (approved):
LGTM

PR #125, opened 2021-09-02
Add more Signals

DasCapschen:
This adds Form Signal Models, as an alternative to the current combination signal.
It also enhances the combination signal with signal boards.

Files were also reorganized a bit. The default signal is now `Ks`.

Btw, a description of all German signals can be found here :)
https://www.bahnstatistik.de/Signale_pdf/SB-DBAG.pdf

TODO:
- [x] add Hp Form Signal
- [x] add Hp Light Signal
- [x] add Vo Form Signal
- [x] add Vo Light Signal
- [x] Add Hl Light Signals (DDR)
- [x] add Hectometer Boards (Milestones)
- [x] add presignal board
- [x] add main signal board
- [x] add combination signal board     
- [x] add speed limit boards (Zs3 form signal)
- [x] add speed limit warning boards (Zs3v form signal)
- [x] add speed limit board Lf7
- [x] add speed limit warning board Lf6
- [x] add presignal warning board (Ne3 form signal)
- [x] add whistle board (Bü4 form signal)
- [x] add tutorials for the new signals

Hp Form Signal:
hp

Vo Form Signal:
vo

Combined H/V Form Signal:
combined_hp_vo

Ks Signal with new Signal Boards:
(the number above the Ks Signal is Zs3 light signal, the number below is Zs3v light signal)
ks

Zs3 and Zs3v + Small Variants:
(Zs3 limits speed until the end of the next switching area ; for example if you go into a station and get Zs3 "8", you need to go 80 until you pass the last switch after the station. Then you're back to line speed. This is a form variant of the number on the Ks signal! Zs3v is the warning for Zs3)
zs3_variants

Lf6 and Lf7 signals:
(Lf7 (white rectangle) limits the *line* speed. Signals or Zs3 can set a *lower* limit, overriding Lf7. Lf6 (yellow triangle) is the warning for Lf7.)
lf_variants

PR #140, opened 2021-09-28
Fix signals and speed limits trying to set themselves to rails, without being in tree

HaSa1002:
fixes the logging issue: https://github.com/Jean28518/Libre-TrainSim/issues/98#issuecomment-928993279


Jean28518:
We also need WarnSpeeds and Stations for this fix. they should work the same way.

Could you add those two please?

HaSa1002:
done

HaSa1002:
Do @DasCapschen the favor and merge #125 first 😆

HaSa1002:
rebased

PR #144, opened 2021-10-02
Remove trailing whitespaces and tabs

HaSa1002:
Godot does that for you if you enable texteditor/files/trim_trailing_whitespace_on_save in the editor settings.

This ensures, we don't clutter our history with unnecessary whtespace changes

PR #133, opened 2021-09-05
Refactor menus

HaSa1002:
The current UI and menu setup is relativly unflexible and hard to maintain.

Apart that, there are a couple of things, I changed or added:
1. A new theme (see below)
2. Pack loading is done in seperate script
4. Settings to the pause menu
5. Automatic font sizing depending on mobile or desktop version
6. Automatic detection of mobile version
7. A loading screen

<details>
<summary>Images</summary>

content menu
about menu
main menu
pause menu

</details>

## Tasks
* [x] Unify mobile and desktop UI where only text size is changed.
* [x] Add AutoDynamicFont changing size depending on device
* [x] Split HUD into mutliple scripts and scenes
* [x] Replace a lot fonts
* [x] Fix the remaining fonts
* [x] Split main menu into multiple scenes and scripts
* [x] Make files snake_case
* [x] Clean up git history
* [x] Finish theme

Fixes #126
~Depends on: #144~

PR #145, opened 2021-10-02
Add better logging

HaSa1002:
* Adds `Logger.log(msg, context=null)` as print replacement
* Adds `Logger.vlog(msg, context=null)` as print verbose option
* Adds `Logger.warn(msg, context)` for warnings
* Adds `Logger.err(msg, context)` for errors

All logger functions log the time and the severity.

Used format:
`[YYYY-MM-DD][HH:MM:SS](context) message`

Depends on #133 and #144

### DIFF
--- a/src/Resources/Basic/SignalTypes/Hl/Hl.gd
+++ b/src/Resources/Basic/SignalTypes/Hl/Hl.gd
@@ -13,7 +13,7 @@ func _ready():
 	signal_logic.signal_type = signal_logic.SignalType.COMBINED
 
 	if signal_logic.speed == -1:
-		printerr(signal_logic.name, ": HL Signals cannot have Speed -1! Please set a speed limit for this Signal! Setting to 160 km/h")
+		Logger.warn("HL Signals cannot have Speed -1! Please set a speed limit for this Signal! Setting to 160 km/h", signal_logic.name)
 		signal_logic.speed = 160
 
 	# Hl signals ALWAYS have a speed limit! NEVER -1 !!!
--- a/src/addons/Libre_Train_Sim_Editor/Data/Modules/UI/loading_screen.gd
+++ b/src/addons/Libre_Train_Sim_Editor/Data/Modules/UI/loading_screen.gd
@@ -45,11 +45,10 @@ func _process(_delta: float) -> void:
 		elif err == OK:
 			update_progress_bar()
 		else: # error during loading
-			printerr("An error occured during loading!");
-			print(err)
+			Logger.err("An error occured during loading! (%s)" % err, self);
 			loader = null
 			OS.shell_open(ProjectSettings.globalize_path("user://logs/"))
-			get_tree().quit(1)
+			get_tree().change_scene("res://addons/Libre_Train_Sim_Editor/Data/Modules/main_menu.tscn")
 			break
 
 
--- a/src/addons/Libre_Train_Sim_Editor/Data/Scripts/ContactPoint.gd
+++ b/src/addons/Libre_Train_Sim_Editor/Data/Scripts/ContactPoint.gd
@@ -37,7 +37,7 @@ func _ready():
 # warning-ignore:unused_argument
 func set_to_rail(newvar):
 	if not find_parent("World"):
-		print("SpeedSign can't find World Parent!'")
+		Logger.err("ContactPoint can't find World Parent!", self)
 		return
 
 	if find_parent("World").has_node("Rails/"+attached_rail) and attached_rail != "":
@@ -83,10 +83,10 @@ func activateContactPoint(trainName):
 func _on_Timer_timeout():
 	var signalN = get_parent().get_node(affectedSignal)
 	if signalN == null:
-		print("Contact Point "+ name + " could not find signal "+affectedSignal+" aborting...")
+		Logger.err("Contact Point "+ name + " could not find signal "+affectedSignal+" aborting...", self)
 		return
 	if signalN.type != "Signal":
-		print("Contact Point "+ name + ": Specified signal point is no Signal. Aborting...")
+		Logger.err("Contact Point "+ name + ": Specified signal point is no Signal. Aborting...", self)
 		return
 	signalN.set_status(newStatus)
 	signalN.set_speed(newSpeed)
--- a/src/addons/Libre_Train_Sim_Editor/Data/Scripts/Signal.gd
+++ b/src/addons/Libre_Train_Sim_Editor/Data/Scripts/Signal.gd
@@ -120,10 +120,10 @@ func _ready():
 # signals necessary for RailMap to work
 func set_status(new_val):
 	if signal_type == SignalType.PRESIGNAL and new_val == SignalStatus.RED:
-		print(name, ": Cannot set a presignal to red!")
+		Logger.err(name + ": Cannot set a presignal to red!", self)
 		return
 	if signal_type == SignalType.MAIN and new_val == SignalStatus.ORANGE:
-		print(name, ": Cannot set a main signal to orange!")
+		Logger.err(name + ": Cannot set a main signal to orange!", self)
 		return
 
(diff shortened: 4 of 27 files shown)


HaSa1002:
Note to self: Use get_stack() for stacktrace

HaSa1002:
rebased and stack trace added. Ready for review

src/Resources/Basic/SignalTypes/Hl/Hl.gd
```diff
@@ -13,7 +13,7 @@ func _ready():
 	signal_logic.signal_type = signal_logic.SignalType.COMBINED
 
 	if signal_logic.speed == -1:
-		printerr(signal_logic.name, ": HL Signals cannot have Speed -1! Please set a speed limit for this Signal! Setting to 160 km/h")
+		Logger.warn(signal_logic.name + ": HL Signals cannot have Speed -1! Please set a speed limit for this Signal! Setting to 160 km/h", self)
```
DasCapschen: do we still need to use `signal_logic.name`, if we're passing `self` as context? Since context is automatically printed.

src/addons/Libre_Train_Sim_Editor/Data/Modules/UI/loading_screen.gd
```diff
@@ -45,11 +45,10 @@ func _process(_delta: float) -> void:
 		elif err == OK:
 			update_progress_bar()
 		else: # error during loading
-			printerr("An error occured during loading!");
-			print(err)
+			Logger.err("An error occured during loading! (%s)" % err, self);
 			loader = null
 			OS.shell_open(ProjectSettings.globalize_path("user://logs/"))
-			get_tree().quit(1)
+			get_tree().change_scene("res://addons/Libre_Train_Sim_Editor/Data/Modules/main_menu.tscn")
```
DasCapschen, line 51: Should this change really be in this PR? Seems like something that was meant to go in #154 ...
I mean, I guess you can just leave it in now, but still 😅

src/addons/Libre_Train_Sim_Editor/Data/Scripts/ContactPoint.gd
```diff
@@ -83,10 +83,10 @@ func activateContactPoint(trainName):
 func _on_Timer_timeout():
 	var signalN = get_parent().get_node(affectedSignal)
 	if signalN == null:
-		print("Contact Point "+ name + " could not find signal "+affectedSignal+" aborting...")
+		Logger.err("Contact Point "+ name + " could not find signal "+affectedSignal+" aborting...", self)
 		return
 	if signalN.type != "Signal":
-		print("Contact Point "+ name + ": Specified signal point is no Signal. Aborting...")
+		Logger.err("Contact Point "+ name + ": Specified signal point is no Signal. Aborting...", self)
```
DasCapschen, line 89: again, is `name` still needed if `self` is context?

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Signal.gd
```diff
@@ -120,10 +120,12 @@ func _ready():
 # signals necessary for RailMap to work
 func set_status(new_val):
 	if signal_type == SignalType.PRESIGNAL and new_val == SignalStatus.RED:
-		print(name, ": Cannot set a presignal to red!")
+		Logger.err(name + ": Cannot set a presignal to red!", self)
+		# assert?
```
DasCapschen: assert would crash the game, we don't want that. The signal code itself does not check if something is a main or presignal, I think, and just tries to set states. 
We would probably fix this if we introduces proper signal blocks...

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Signal.gd
```diff
@@ -149,7 +151,7 @@ func set_to_rail(newvar):
 	if !is_inside_tree():
 		return
 	if world == null:
-		print(name, ": CAN'T FIND WORLD NODE!")
+		Logger.err(name + ": CAN'T FIND WORLD NODE!", self)
```
DasCapschen, line 152: again, `name` necessary?

src/addons/Libre_Train_Sim_Editor/Data/Scripts/editor_configuration.gd
```diff
@@ -25,6 +25,8 @@ func initialize_UI():
 
 func initialize_editor_directory():
 	var dir = Directory.new()
+	if dir.open("user://") != OK:
+		Logger.err("Can't open directory '%s'" % editor_directory, self)
```
DasCapschen, line 29: return?

src/addons/Libre_Train_Sim_Editor/Data/Scripts/editor_configuration.gd
```diff
@@ -35,8 +37,11 @@ func _on_UpdateEditorPathButton_pressed():
 	initialize_editor_directory()
 	jSaveManager.save_setting("editor_directory_path", editor_directory)
 
+
 func initialize_track_directory(entry_name):
 	var dir = Directory.new()
+	if dir.open(editor_directory) != OK:
+		Logger.err("Can't open directory '%s'" % editor_directory, self)
```
DasCapschen, line 45: return?

src/addons/Libre_Train_Sim_Editor/Data/Scripts/editor_configuration.gd
```diff
@@ -49,6 +54,8 @@ func initialize_track_directory(entry_name):
 
 func create_resource_directory_structure(track_name):
 	var dir = Directory.new()
+	if dir.open(editor_directory) != OK:
+		Logger.err("Can't open directory '%s'" % editor_directory, self)
```
DasCapschen, line 61: return?

src/addons/Libre_Train_Sim_Editor/Data/Scripts/editor_configuration.gd
```diff
@@ -64,34 +71,40 @@ func _on_TracksList_user_added_entry(entry_name):
...
 	new_name = remove_inappropriate_signs_from_track_name(new_name)
-
-	jEssentials.copy_folder_recursively(editor_directory + "Worlds/" + old_name, editor_directory + "Worlds/" + new_name)
-
+	jEssentials.copy_folder_recursively(editor_directory + "Worlds/" + old_name, \
+			editor_directory + "Worlds/" + new_name)
 	var dir = Directory.new()
-	if dir.open(editor_directory + "Worlds/" + new_name) != OK: return
+	if dir.open(editor_directory + "Worlds/" + new_name) != OK:
+		Logger.err("Can't open directory '%s'" % editor_directory, self)
```
DasCapschen, line 96: return?

DasCapschen (changes requested):
Looks great, but please check comments. 👍🏻

src/addons/Libre_Train_Sim_Editor/Data/Modules/UI/loading_screen.gd
HaSa1002: Yeah. I guess I changed that because of a merge conflict while rebasing

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Signal.gd
HaSa1002: assert does not crash the game. It causes a debug break in Debug mode, but is removed in relase builds. That is the whole point of asserts :D

DasCapschen: oh, it's removed in release builds? Well, okay. Still, we should not assert, because you would get debug breaks all the time. The signal system just doesn't care, and imo it doesn't have to.

DasCapschen (approved):
lgtm

PR #153, opened 2021-10-03
Fix trains embedded meshes and wrong Wheel direction

DasCapschen:
JFR1_Red and JFR1_Grey had some meshes (Cabin, Brake_Roll, Acc_Roll, Pantograph) embedded into the scene instead of loading them from the external resource obj. This caused those files to be very big.
Fixed in this PR.

Also fixes the wheels of the train, which were facing the wrong way 🙈


HaSa1002:
Activity farming xD
```
+24,043 −25,330
```

DasCapschen:
rebased.

HaSa1002 (approved):
lgtm

PR #155, opened 2021-10-03
Fix save paths

### LINKED ISSUE #142: Save path invalid
The game requests the save path to be in `/home`, which is not only a bad path, but also not a valid path on Windows. We should move the path to `user://`

HaSa1002:
Should fix all invalid directory errors
Fixes #142


DasCapschen (commented):
looks good 👍🏻

PR #156, opened 2021-10-03
fix normals, enable backface culling

DasCapschen:
Many materials had back-face culling disabled because some meshes (the train Door, Cabin and Control Roll) had wrong normals. 
I have fixed the normals and enabled back-face culling for the materials again.

There is a new material for the overhead line, because that one actually needs culling disabled :)

I didn't see any problems with missing faces on other meshes, but if you spot anything, let me know and I can fix those too.


Jean28518:
Thanks for cleaning up this. learning curve was was very steep at beginning

PR #158, opened 2021-10-03
Add CI

### LINKED ISSUE #85: Automation of building of Libre TrainSim?

HaSa1002:
No Android support, because we need to set up SECRETS for them.
Linux fails because of memory leaks.
I would like to add adhoc notarization in the future for mac as that would help mac users to actually use these builds.

@Jean28518 you need to manually approve the workflow run.

For an example visit: https://github.com/HaSa1002/Libre-TrainSim/pull/1
Don't mind the ton of commits there. I haven't updated my main branch for a while

Resolves #85

### DIFF
--- a/.github/workflows/export.yml
+++ b/.github/workflows/export.yml
@@ -1,0 +1,213 @@
+name: CI
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+env:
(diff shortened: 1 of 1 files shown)


HaSa1002:
Linux works once #159 is merged

Jean28518:
Very interesting, need to learn github workflows ;)

PR #162, opened 2021-10-08
Overhaul JFR1 Train

DasCapschen:
The train was using 17 materials, which was a bit much, so I baked the materials into a single texture. The train now uses 3 materials: Glass, Lamp, and the Rest :)
This should result in better performance, because there are less draw calls (and maybe faster loading?)
Textures are 4k (4096x4096) and include: Albedo, Metal, Roughness and Ambient Occlusion (AO, Metal, Roughness are combined into 1 "AMR" texture, because they are all just grey images)

I also added lights to the inside of the Wagons and Cabin for better visibility, because the AO texture made them darker.
We might want to bake the lights instead of using them real-time so they use up less performance.

I also changed the Wagon script because I moved Doors, Seats and PathNodes into sub-nodes for easier overview when editing Wagons.

Depends on: #153


HaSa1002:
Wtf. the noise is heavy
grafik

DasCapschen:
The noise is either: the texture itself, it is not perfectly smooth (uses the Plastic.png texture). But it looks worse because VRAM compression...
Or it's the AO map. I cannot denoise the AO map though, because then you will complain about banding 😅
And the banding is super noticable ingame, more than the noise I would say. And there's nothing we can do about it, because the grays are literally a single bit apart.
We could just disable AO...

Noisy AO:
Cabin_ao-0
Denoised, banding AO:
Cabin_ao-0-denoised

HaSa1002 (approved):
noise is fixed in the cabin. All in all a nice improvement. LGTM.

PR #163, opened 2021-10-10
Add: PZB Safety System

### LINKED ISSUE #131: Add PZB
For more realistic behaviour, we should implement more security systems, like PZB.

I would like to implement this myself, leaving this issue here as a TODO / Documentation of how PZB works.

I recommend these sources for explanations :)
https://de.wikipedia.org/wiki/Punktf%C3%B6rmige_Zugbeeinflussung
https://www.youtube.com/watch?v=M5cfjwSpIds
http://tf-ausbildung.de/BahnInfo/oberezugart.htm

# For passenger trains (pzb type 85): 
## Signals
- add 1000 Hz magnet on pre-signals
- triggers PZB warning if next main signal is Red or has a speed limit
     + 85 lamp starts blinking, 1000 Hz lamp illuminates
- driver must press "PZB Acknowledge" button within 4 seconds (else train emergency brakes)
- slow down to 85km/h within 23 seconds (else emergency brakes)
- after 700m, 1000 Hz mode turns off. (1000 Hz lamp turns off)
- If signal is green, press "PZB Free" button to exit monitoring (85 lamp stops blinking)
     + continue your journey like normal
     + if "PZB free" was pressed, but 1000 Hz lamp is on, or signal is red: emergency braking after passing 500 Hz magnet
- If signal is red, slow down to 65km/h *before* the 500 Hz magnet
- after passing 500 Hz magnet (500Hz lamp turns on), slow down to 45km/h within 153 meters
    + you *cannot* use "PZB free" if 500 Hz mode is on! Use it *before* passing the 500 Hz magnet!
- if signal is green: stay below 45km/h and pass the signal
- if signal is red: stop before the signal (2000 Hz magnet) and wait until it is green, else emergency brakes
    + you can press and hold the "PZB command" button while driving over the 2000 Hz magnet / red signal without getting emergency brakes (up to 40km/h only!) (if you're allowed to :wink: by a replacement signal or otherwise)
- accelerate to pass the signal. 250m *after* the signal, monitoring turns off. (do not go over 45km/h within those 250 meters!)
- Continue your journey like normal.

## Restrictive Mode
If at any point, the train stopped or is going below 10km/h for more than 15 seconds:
- enter restrictive monitoring (70 and 85 lamp blink alternating)
- limit train to 45km/h (1000 Hz mode / train starts for the first time) or 25km/h (500 Hz mode)
- press "PZB Free" button to release from restrictive monitoring (this does *not* work if 500 Hz mode is on!)
- if train was released:
     + if 1000 Hz magnet is passed, go to *normal* 1000 Hz mode (not restrictive)
     + if 500 Hz magnet is passed, emergency brakes come on

DasCapschen:
This PR does 2 things:
- refactor how safety systems work (so far this was only sifa)
- add a second safety system, PZB

Safety systems are now separate nodes, placed in the train scene. Their script checks whatever the safety system should do, and if it detects an issue, it will request emergency braking of the train. 
Because of this change, Player.gd has no Sifa functionality anymore, and the `check_security()` function has changed. The Player will now iterate over all safety systems and check if they requested emergency braking. If any system did, the train will brake.

For a description of PZB, see #131 

This PR closes #131


src/addons/Libre_Train_Sim_Editor/Data/Modules/PZBModule.tscn
```diff
@@ -0,0 +1,253 @@
...
+	else:
+		$PanelContainer/GridContainer/Monitor700m.text = str($\"../700mMonitor\".distance_left())
+	
+	if $\"../1250mMonitor\".is_stopped:
+		$PanelContainer/GridContainer/Monitor1250m.text = \"STOPPED\"
+	else:
+		$PanelContainer/GridContainer/Monitor1250m.text = str($\"../1250mMonitor\".distance_left())
+
+	$PanelContainer/GridContainer/Brakes.text = str(pzb.requires_emergency_braking)
+"
```
HaSa1002: embedded GDScript. Good or bad?

src/addons/Libre_Train_Sim_Editor/Data/Scripts/DistanceMonitor.gd
```diff
@@ -0,0 +1,40 @@
+extends Node
+class_name DistanceMonitor
+
+export (int) var _distance = 0
```
HaSa1002:
```suggestion
export var _distance: int = 0
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/DistanceMonitor.gd
```diff
@@ -0,0 +1,40 @@
...
+var _player = null
+var _start_dist = 0
+
+var is_stopped = true
+
+signal reached()
```
HaSa1002: signal definitions should be the first thing at the top

src/addons/Libre_Train_Sim_Editor/Data/Scripts/DistanceMonitor.gd
```diff
@@ -0,0 +1,40 @@
...
+
+func set_distance(val):
+	self._distance = val
+
```
HaSa1002, line 15: there should be two empty lines between functions

src/addons/Libre_Train_Sim_Editor/Data/Scripts/SifaModule.gd
```diff
@@ -0,0 +1,67 @@
...
+var is_sifa_enabled = true
+var was_sifa_reset = false
+
+# trigger a lamp in the trains cabin
+# this is a signal, so different trains can implement it differently
+signal sifa_visual_hint(is_turned_on)
+
+var player
+func _ready():
+	player = find_parent("Player")
```
HaSa1002:
```suggestion
onready var player = find_parent("Player")
func _ready():
```
and maybe we should put that into SafetySystem

src/addons/Libre_Train_Sim_Editor/Data/Scripts/TrainInfo.gd
```diff
@@ -82,19 +82,19 @@ func update_info(player):
...
-		$ScrollContainer/VBoxContainer/SiFa/dot.texture = orange
-	else:
-		$ScrollContainer/VBoxContainer/SiFa/dot.texture = green
+#	$ScrollContainer/VBoxContainer/SiFa.visible = player.sifaEnabled
+#	if player.sifaTimer > 35:
+#		$ScrollContainer/VBoxContainer/SiFa/dot.texture = red
+#	elif player.sifaTimer > 25:
+#		$ScrollContainer/VBoxContainer/SiFa/dot.texture = orange
+#	else:
+#		$ScrollContainer/VBoxContainer/SiFa/dot.texture = green
```
HaSa1002:
```suggestion
```
YAGNI?

HaSa1002 (changes requested):
just minor stuff and code style

src/addons/Libre_Train_Sim_Editor/Data/Modules/PZBModule.tscn
DasCapschen: I guess it shouldn't be embedded. That's just the script of the debug panel, good catch, I forgot to hide that by default 😅

src/addons/Libre_Train_Sim_Editor/Data/Scripts/SifaModule.gd
DasCapschen: I'm not sure, I feel like it shouldn't go into SafetySystem.

src/addons/Libre_Train_Sim_Editor/Data/Modules/PZBModule.tscn
HaSa1002: FYI there are a few more embedded

DasCapschen: yeah, sifa was also embedded, but I fixed that. That should've been all though.

src/addons/Libre_Train_Sim_Editor/Data/Scripts/SifaModule.gd
HaSa1002: fair

HaSa1002 (approved):
lgtm

HaSa1002:
1. The train does not work
grafik
2. Does the key change depending on the action?
grafik
3. Is there a message informing the player? Do we want that?
4. Do the magnets have textures?
5. Should Sifa be enabled in easy mode?
6. "Inavlid set index overrunnedRedSignal with value of type bool" (and it should be overrunRedSignal)

DasCapschen:
The train should've been fixed with #153 and #162 🤔 ... maybe I messed up a merge there.

Thanks for catching the keys in the text, I was meaning to code that, but then forgot... so no, the key text is hardcoded 🙈 

Informing the player? Of what?

The magnets are currently 2 or 3 materials I think, I can bake it into textures, probably better that way.

No, Sifa and PZB should *not* be enabled in easy mode, and they should also not be enabled in the mobile version. (since mobile is pretty much the easy mode)
We should make extra difficulty settings for enabling sifa / pzb separately.

HaSa1002:
> Informing the player? Of what?

I mean sending a message when you are not pressing PgDn.

> The magnets are currently 2 or 3 materials I think, I can bake it into textures, probably better that way.

I asked, because the magnet looked plain white.

> No, Sifa and PZB should not be enabled in easy mode, and they should also not be enabled in the mobile version. (since mobile is pretty much the easy mode)

Sifa should be disabled then (It isn't right now).

DasCapschen:
> I mean sending a message when you are not pressing PgDn.

No, we shouldn't. Real PZB doesn't do it. I would, by default, disable PZB (we really need difficulty settings) and let the player decide if they are ready for it. 


> I asked, because the magnet looked plain white.

Then you've been trying a rather old version, I fixed that a while ago 😅 
Or did I forget to push? :thinking:

> Sifa should be disabled then (It isn't right now).

will do :+1:

HaSa1002 (approved):
lgtm

HaSa1002:
thanks :)

PR #164, opened 2021-10-10
add drag and drop for rails in editor :)

### LINKED ISSUE #151: Add drag and drop support for track building
It would improve the usability of the editor immensely, if dragging of tracks or their contact points are implemented.
This way you drag a contact point and move it with your mouse to destination and the track would update realtime to approximatly match the desired target point. Refer to the track building in Transport Fever for inspiration.

Understanding the track building as part of our gameplay would improve the way people interact with the system. They would have fun using the editor and thus like to create own tracks

DasCapschen:
Fixed #151 😄 

don't merge just yet, there might be some bugs.
we might want to add a minimum radius.


src/addons/Libre_Train_Sim_Editor/Data/Scripts/Editor.gd
```diff
@@ -49,9 +47,15 @@ func _exit_tree():
 	Root.Editor = false
 
 func _input(event):
-	if event is InputEventMouseButton and event.button_index == BUTTON_LEFT and event.pressed == false and not $EditorHUD.mouse_over_ui:
+	if event is InputEventMouseButton and event.button_index == BUTTON_LEFT and event.pressed == true and not $EditorHUD.mouse_over_ui:
```
HaSa1002:
```suggestion
	if event is InputEventMouseButton and event.button_index == BUTTON_LEFT and event.pressed and not $EditorHUD.mouse_over_ui:
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Editor.gd
```diff
@@ -49,9 +47,15 @@ func _exit_tree():
...
 		select_object_under_mouse()
 	
+	if event is InputEventMouseButton and event.pressed == false and drag_mode:
```
HaSa1002:
```suggestion
	if event is InputEventMouseButton and !event.pressed and drag_mode:
```

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Editor.gd
```diff
@@ -62,9 +66,26 @@ func _input(event):
...
+	drag_mode = false
+	#print("DRAG MODE END")
+
+
+func handle_drag_mode():
+	var mouse_pos = get_viewport().get_mouse_position()
+	var plane = Plane(Vector3(0,1,0), selected_object.startpos.y)
+	var mouse_pos_3d = plane.intersects_ray($FreeCamera.project_ray_origin(mouse_pos), $FreeCamera.project_ray_normal(mouse_pos))
+	if mouse_pos_3d != null:
+		selected_object.calculate_from_start_end(mouse_pos_3d)
```
HaSa1002: Could be done with PhysicsObject with InputEvent tracking.
grafik
using those callbacks
grafik

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Rail.gd
```diff
@@ -343,6 +343,34 @@ func get_shifted_local_pos_at_RailDistance(distance, shift):
...
+	
+	if abs(end.z) < 0.01:
+		radius = 0
+		length = end.length()
+		update()
+		return
+	
+	# m = end.z / end.x
+	# m2 = - 1 / m
+	# b = z - m2 * x
```
HaSa1002, line 359:
```suggestion
```
YAGNI

HaSa1002 (changes requested):

HaSa1002:
Cursor is offset.
Repro:
1. Create Rail with curvature
2. Add new rail using "after rail"
3. Drag the endpoint of the second rail

HaSa1002:
if you drag the endpoint onto the movement gizmo, you trigger the movement

HaSa1002:
**Issue is not related to PR**
power line is not updated correctly
grafik

HaSa1002:
I would favor if you could change the addition behaviour to (but not a blocker and can be made in a follow-up PR):
1. Tell game, you want to add a rail
2. Move your mouse to start position
3. Press LMB and hold
	1. If near other rail, snap to that rail and get curvature and radius from that rails
	2. Else set Rotation to 0
4. Drag rail
5. Release LMB to end action

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Editor.gd
DasCapschen: Not sure if we should use a node here. Because the y position should be the height of the current rail, not of a fixed collision plane node.
Though in the future, we would want collision with a heightmap shape there, so that might be useful afterall... 🤔

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Rail.gd
DasCapschen: was for documentation purposes, not as commented code

DasCapschen:
ty for review
- style changes applied
- cursor offset fixed
- ui updates while dragging
- gizmo hidden while dragging (gizmo itself still broken)
- powerline is a Rail.gd issue
- gonna code snapping now :)

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Editor.gd
HaSa1002: I would create a sphere shape as collider and use the input_event callback for the xz movement. Raycast can then be used to determine the height (or just statically set to value for now)

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Editor.gd
```diff
@@ -62,22 +68,56 @@ func _input(event):
...
+func handle_drag_mode():
+	var mouse_pos = get_viewport().get_mouse_position()
+	var plane = Plane(Vector3(0,1,0), selected_object.startpos.y)
+	var mouse_pos_3d = plane.intersects_ray($FreeCamera.project_ray_origin(mouse_pos), $FreeCamera.project_ray_normal(mouse_pos))
+	if mouse_pos_3d != null:
+		selected_object.calculate_from_start_end(mouse_pos_3d)  # update rail
+		provide_settings_for_selected_object()  # update ui
 	
+	# wait for physics update, else we can never un-snap
+	yield(get_tree(), "idle_frame")
```
HaSa1002:
```suggestion
	# wait for physics update, else we can never un-snap
	yield(get_tree(), "physics_frame")
```
Physics frame is `physics_frame`

HaSa1002 (changes requested):

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Editor.gd
DasCapschen: good point 🙈

DasCapschen: actually, no, this needs to be idle_frame, it doesn't work otherwise... I'm not quite sure why 🤔 
basically the problem is that the position of the area needs to update before we check the snapping, else it will stay snapped once it has snapped.

HaSa1002: okay. It works. Maybe remove the comment though

src/addons/Libre_Train_Sim_Editor/Data/Scripts/Editor.gd
DasCapschen: But also, that would require connecting to the signal and installing a callback, right?
The way it is now, the code runs independent of that, without connecting and disconnecting signals.
I'm just not sure if it's really necessary to use those signals, they don't seem to provide any benefit over the code we have now.

HaSa1002:
I doesn't do anything after I press OK and instead get that error:
```
ERROR: Signal 'confirmed' is already connected to given method '_snap_complex_connector' in that object.
   At: core/object.cpp:1512
```

HaSa1002:
and when it starts to show the text box, it says:
```
ERROR: Signal 'confirmed' is already connected to given method '_snap_complex_connector' in that object.
   At: core/object.cpp:1512
[]
ERROR: Signal 'confirmed' is already connected to given method '_snap_complex_connector' in that object.
   At: core/object.cpp:1512
[]
ERROR: Signal 'confirmed' is already connected to given method '_snap_complex_connector' in that object.
   At: core/object.cpp:1512
[]
ERROR: Signal 'confirmed' is already connected to given method '_snap_complex_connector' in that object.
   At: core/object.cpp:1512
[]
ERROR: Signal 'confirmed' is already connected to given method '_snap_complex_connector' in that object.
   At: core/object.cpp:1512
```

DasCapschen:
Looks like you're triggering the complex connector, for arbitrary connections. That's not implemented yet, I guess I should have disabled the popup for it, but as it stands, this PR is WIP.

Signal errors should be fixed with a simple if check. (though I kinda wish godot just wouldn't throw an error on double connect...)
I have no idea where the `[]` print comes from...

HaSa1002:
Yeah I figured the signals will be connected multiple times, because the snapping is in place, but the the popup connection logic is triggered every snapping mouse movement frame.

That is totally fine. I was just super confused, but I found the two working use cases, so no worried there (although the complex one may be used even in the 90° case).

HaSa1002:
🥳

PR #177, opened 2021-10-16
Fix camera in editor

HaSa1002:
Fixes wrong camera being used in handle_drag_mode and fixes camera rotation teleporting after load or station jump.


DasCapschen (approved):
lgtm

PR #189, opened 2021-10-17
Fix select collider not working with signals

### LINKED ISSUE #179: Select Collider is not working (Signal)
**Stacktrace:**
0 - SelectCollider.gd:9:_ready
1 - Editor.gd:584:add_signal_to_selected_rail
2 - AddObjects.gd:57:_on_RailLogicMenu_item_selected

**Error:**
Attempt to call function `get_node` in base `null instance` on a null instance

HaSa1002:
It is not the select collider who is the problem, but the signal that creates the visual instance not in `_ready`. It will be create eventually through a series of calls started by the timer.

The select collider itself is not a beautiful either and should be reworked, but for now it works again.

Fixes #179

### DIFF
--- a/src/addons/Libre_Train_Sim_Editor/Data/Scripts/Signal.gd
+++ b/src/addons/Libre_Train_Sim_Editor/Data/Scripts/Signal.gd
@@ -1,4 +1,3 @@
-tool
 extends Spatial
 const type = "Signal" # Never change this type!!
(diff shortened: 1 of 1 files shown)


Jean28518 (approved):

DasCapschen (approved):
lgtm

PR #190, opened 2021-10-17
Fix jList copy paste

HaSa1002:
This was broken because of the UI refactoring. We need to put a string into the clipboard. This is safe (it doesn't allow objects; only vars)

### DIFF
--- a/src/addons/jean28518.jTools/jList/jList.gd
+++ b/src/addons/jean28518.jTools/jList/jList.gd
@@ -249,11 +249,11 @@ func _on_Copy_pressed(): # stores the current entry_names into the global buffer
 	var source_entry_ids = item_list.get_selected_items()
 	for entry_id in source_entry_ids:
 		source_entry_names.append(item_list.get_item_text(entry_id))
-	OS.clipboard = source_entry_names
+	OS.clipboard = var2str(source_entry_names)
 	emit_signal("user_copied_entries", source_entry_names)
 
 func _on_Paste_pressed(): # Adds entry_names from global buffer into jList.
-	var source_entry_names = OS.clipboard
+	var source_entry_names = str2var(OS.clipboard)
 	if source_entry_names == null:
 		return
 	undo_buffer = get_data()


DasCapschen:
Do we really want to use the OS clipboard? Why not just a buffer?

Jean28518:
I don't see any advantage in using the OS.clipboard.

HaSa1002:
We don't need to have registration code and thus don't need a singleton that manages the jLists. That was a ton of boiler plate code for just copy paste. Using the clipboard is the easiest way. It is not unsafe either as no objects can be created in str2var. We cannot not simply use a buffer, because the jList is also able to copy pase between multiple lists

DasCapschen:
Well that's exactly what a Singleton would be good for.
You copy something, which means writing it to some global buffer, and when pasting you just take what's in the buffer and write it to your goal.
If you want, we can use str2var() and var2str() and let the buffer *be* a string, but I see no reason to pollute the users clipboard with this.

DasCapschen:
I've talked to hasa and looked at the code, and we really are just copying text. Even copying objects just copies their name and pasting then resolves to get the data.
As long as copying stays this way, i agree with using OS.Clipboard
But I don't ever want to have binary data or serialized objects in my clipboard, so if we ever extend or rework copy-paste, we need to pay attention to this.

DasCapschen:
If, for whatever reason, we need serialised object dumps, or binary data from the editor, we should just add an extra "dump to clipboard" button, but not do this through default copy paste

DasCapschen:
What do you think, @Jean28518 ?

Jean28518:
Why do we want to pollute the users clipboard? I don't see any advantage for the user.

I am open to other solutions in Libre TrainSim (Godot) but in my opinion we should stay in the Godot ecosystem at this point and not use any "external" things, unless we really have to..

HaSa1002:
The benefit of using the clipboard is consistent copy and paste. I, as the user, tell the engine to copy the data. It is not like we are doing the copy command "from code" because we feel like there should be a copy being made. It is the user who is requesting a copy of the data. IMO that is reason why it should be OS.clipboard.

Jean28518:
While we are discussing, this is already upstream......

grafik

Don't have the energy to discuss that further on

PR #196, opened 2021-10-17
Prevent scripts from running in editor

HaSa1002:
We don't need tool scripts anymore as the editor is running in game.
These tool scripts do either nothing, rise complexity, or cause some
bugs. We simply don't need them. YAGNI


Jean28518 (approved):
Is okay!

DasCapschen (approved):
lgtm

PR #204, opened 2021-10-23
performance hotfix

DasCapschen:
Fixes bad performance caused by too many lights used in the wagons.

Adds an option "dynamic lights" in the options menu. 
When turned *off* (default) it will delete the wagon lights on `_ready`.

This should be considered only a hotfix so we can release 0.9 without performance problems, but we should really rework this in 0.10.

I heard @HaSa1002 already has an idea for this? :smile:


HaSa1002 (approved):
I don't like this fix. It is code wise ok, but it has implications on the way trains need to be made. I'll approve the change, but I am *not* in favor of merging that fix. If I find time to make a better fix, I would like to supersede this PR instead. If I don't find the time, we can merge this. It's not like the trains have clean code anyway.

TL;DR: Please wait with merging, but approved in case we don't find the time for fixing it properly before 0.9

PR #208, opened 2021-10-31
Add Highlight postprocessing to Editor selection (remove bouncing animation)

DasCapschen:
This removes the "bouncing" animation from the selected object in Editor, and replaces it with Postprocessing Effects to outline and highlight the object.

This is what it would currently look like:
image

Of course, we can decide to only use outlines, or only use highlighting (they are 2 separate nodes that added to the Editor scene tree), and the colors are also configurable.

What is currently not configurable, though I'm sure I can work out how to do that, is the thickness of the outline, which is currently only about 2 pixels...

### DIFF
--- a/src/Data/Postprocess/HighlightPP.tscn
+++ b/src/Data/Postprocess/HighlightPP.tscn
@@ -1,0 +1,33 @@
+[gd_scene load_steps=4 format=2]
+
+[ext_resource path="res://Data/Postprocess/highlight_shader.gdshader" type="Shader" id=1]
+[ext_resource path="res://Data/Postprocess/ShaderPP.gd" type="Script" id=2]
+
+[sub_resource type="ShaderMaterial" id=1]
+shader = ExtResource( 1 )
+shader_param/highlight_color = Color( 1, 0.913725, 0, 0.231373 )
+
+[node name="HighlightPP" type="Node"]
+script = ExtResource( 2 )
+
+[node name="HighlightShader" type="ViewportContainer" parent="."]
+material = SubResource( 1 )
+anchor_right = 1.0
+anchor_bottom = 1.0
+mouse_filter = 2
+stretch = true
+__meta__ = {
+"_edit_use_anchors_": false
+}
+
+[node name="Viewport" type="Viewport" parent="HighlightShader"]
+size = Vector2( 800, 600 )
+transparent_bg = true
+handle_input_locally = false
+keep_3d_linear = true
+debug_draw = 1
+render_target_update_mode = 3
+
+[node name="Camera" type="Camera" parent="HighlightShader/Viewport"]
+cull_mask = 2
+far = 1000.0
--- a/src/Data/Postprocess/LaplaceEdgeDetect.gdshader
+++ b/src/Data/Postprocess/LaplaceEdgeDetect.gdshader
@@ -1,0 +1,25 @@
+shader_type canvas_item;
+
+uniform vec4 highlight_color: hint_color = vec4(1.0, 0.0, 0.0, 1.0);
+
+void fragment() {
+	vec4 laplace;
+	
+	// laplace edge detection (faster to calc than sobel!)
+	laplace  =  8.0 * texture(TEXTURE, UV);
+	laplace += -1.0 * texture(TEXTURE, UV + vec2(-SCREEN_PIXEL_SIZE.x, -SCREEN_PIXEL_SIZE.y));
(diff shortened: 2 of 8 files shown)


HaSa1002:
I'd be in favor of only using the outline shader because the tint makes it hard to match colors.

src/addons/Libre_Train_Sim_Editor/Data/Modules/Postprocess/LaplaceEdgeDetect.gdshader
```diff
@@ -0,0 +1,38 @@
...
+// converts the texture to a binary texture
+// which means only black or white, depending on if a pixel contains information
+// or not. 
+float tex_bin(sampler2D tex, vec2 uv) {
+	vec3 rgb = texture(tex, uv).xyz;
+	if((rgb) == vec3(0)) {
+		return 0.0;
+	} else {
+		return 1.0;
+	}
```
HaSa1002: `step`?

src/addons/Libre_Train_Sim_Editor/Data/Modules/Postprocess/ShaderPP.gd
```diff
@@ -0,0 +1,31 @@
...
+# check ViewportContainer -> Material for the Shader
+
+export var highlight_color: Color
+
+var external_camera
+var shader_camera
+
+func _ready():
+	external_camera = get_viewport().get_camera()
+	shader_camera = find_node("Camera")
```
HaSa1002:
```suggestion
onready var external_camera := get_viewport().get_camera()
onready var shader_camera := find_node("Camera") as Camera
```

src/addons/Libre_Train_Sim_Editor/Data/Modules/Postprocess/ShaderPP.gd
```diff
@@ -0,0 +1,31 @@
...
+
+func _process(delta: float) -> void:
```
HaSa1002:
```suggestion
func _process(_delta: float) -> void:
```
Surpresses unused argument warning

src/addons/Libre_Train_Sim_Editor/Data/Modules/Postprocess/highlight_shader.gdshader
```diff
@@ -0,0 +1,23 @@
...
+// converts the texture to a binary texture
+// which means only black or white, depending on if a pixel contains information
+// or not. 
+int tex_bin(sampler2D tex, vec2 uv) {
+	vec3 rgb = texture(tex, uv).xyz;
+	if((rgb) == vec3(0)) {
+		return 0;
+	} else {
+		return 1;
+	}
```
HaSa1002: `step`?

HaSa1002 (changes requested):

src/addons/Libre_Train_Sim_Editor/Data/Modules/Postprocess/LaplaceEdgeDetect.gdshader
DasCapschen: what?

HaSa1002:
```suggestion
	return step(texture(tex, uv).xyz, vec3(0)).x;
```

HaSa1002: because step is GPU accelerated it is faster (in theory)

DasCapschen: step is not going to work. It is defined as `b < a ? 0 : 1` and the minimum value rgb can have *is* 0, so `0 < 0` is always false and this will always return 1...

DasCapschen:
I've changed it to only use outlines. 
And I fixed the outline shader... it was discarding half of the outline 🙈 
So they are bigger now too, should be well visible.

src/addons/Libre_Train_Sim_Editor/Data/Modules/Postprocess/LaplaceEdgeDetect.gdshader
HaSa1002: yes. you're right

DasCapschen:
rebased

HaSa1002 (approved):
I mean it is ugly, but who cares xD
grafik

HaSa1002:
thank you

PR #214, opened 2021-11-05
fix world loading

DasCapschen:


src/project.godot
```diff
@@ -208,10 +208,6 @@ ContentLoader="*res://Data/Scripts/Singletons/content_loader.gd"
 LoadingScreen="*res://Data/UI/loading_screen.tscn"
 InputHelper="*res://Data/Scripts/Singletons/input_helper.gd"
 
-[debug]
-
-settings/stdout/verbose_stdout=true
-
```
HaSa1002: 🤨

HaSa1002 (approved):

src/project.godot
DasCapschen: fixed.

HaSa1002: I know you hate that option. We should be able to get rid of that soon-ish

HaSa1002:
thank you

PR #216, opened 2021-11-05
hide TrackObject tab when a non-rail is selected

### LINKED ISSUE #212: It is possible add TrackObjects to objects that are not Rails
When you select an object, and go to the TrackObject Tab, it will let you add an object, but then crash because you did not select  a rail

DasCapschen:
fixes #212

### DIFF
--- a/src/Editor/Modules/Content_Selector.gd
+++ b/src/Editor/Modules/Content_Selector.gd
@@ -39,7 +39,7 @@ func crawl_directory_for_resources():
 
 	if current_type == OBJECTS:
 		for folder in ContentLoader.repo.object_folders:
-			Root.crawlDirectory(folder, current_resources, ["gltf", "glb", "obj", "tscn"])
+			Root.crawlDirectory(folder, current_resources, ["obj"])
 
 	if current_type == RAIL_TYPES:
 		for folder in ContentLoader.repo.rail_type_folders:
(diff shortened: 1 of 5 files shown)


src/Editor/Modules/Content_Selector.gd
```diff
@@ -39,7 +39,7 @@ func crawl_directory_for_resources():
 
 	if current_type == OBJECTS:
 		for folder in ContentLoader.repo.object_folders:
-			Root.crawlDirectory(folder, current_resources, ["gltf", "glb", "obj", "tscn"])
+			Root.crawlDirectory(folder, current_resources, ["obj"])
```
DasCapschen, line 42: yes, this change is actually necessary, because the editor expects objects to be Meshes.

gltf, glb and tscn are PackedScenes though, and not meshes, and will crash the editor

HaSa1002: We should change that at some point

src/Editor/Modules/RailLogic.gd
```diff
@@ -1,15 +1,9 @@
 extends VBoxContainer
 
 
-# Declare member variables here. Examples:
-# var a = 2
-# var b = "text"
 func _input(event):
 	visible = is_instance_valid(current_rail_logic) and get_parent().current_tab == 2
```
HaSa1002, line 5: defaq? Why in _input?

HaSa1002 (approved):
lgtm

PR #217, opened 2021-11-05
automatically add poles when rail is electric

### LINKED ISSUE #167: Power line is not correctly updated in editor
It is just a straight line
grafik

## To Do
- [x] When adding an electric rail, some standard poles should be added automatically
- [ ] This Option should be toggable (default: true)

DasCapschen:
fixes #167

### DIFF
--- a/src/Editor/Docks/RailBuilder/RailBuilder.gd
+++ b/src/Editor/Docks/RailBuilder/RailBuilder.gd
@@ -1,12 +1,12 @@
 extends Control
 
+onready var editor: Node = find_parent("Editor")
 var world: Node
 var currentRail: Node
 var eds # Editor Selection
 
 
 func _process(delta: float) -> void:
-	var editor: Node = find_parent("Editor")
 	if editor:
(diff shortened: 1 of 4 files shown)


src/Editor/Docks/RailBuilder/RailBuilder.gd
```diff
@@ -1,12 +1,13 @@
...
 
+var editor: Node
 var world: Node
 var currentRail: Node
 var eds # Editor Selection
 
 
 func _process(delta: float) -> void:
-	var editor: Node = find_parent("Editor")
+	editor = find_parent("Editor")
```
HaSa1002: do in `onready`?

HaSa1002 (approved):
lgtm. Check comment please

HaSa1002 (changes requested):
Doesn't work when you add a new rail using the new rail button in the context menu of the rails

HaSa1002 (approved):
lgtm

PR #226, opened 2021-11-06
Nicer modding tools

DasCapschen:
I didn't like the way the little modding tools addon I wrote worked, so I changed it.
Now you get a dock next to the file system to create a new mods, and it has a button that links to modding help.
grafik


HaSa1002 (approved):
lgtm

PR #236, opened 2021-11-07
Convert chunk system to use TSCN

### LINKED ISSUE #150: Use scenes and resources to store chunks and scenarios
Right now chunks are stored in a map_name.save file, which needs to be manually added to the games export includes.
Furthermore it requires loading the world in order to start populating it with content. This way the loading screen do have some functionality, but its main goal to show sensible progress can't be met. It would be easier, if we store scenearios as resources, the chunks as packedscenes and then load the world, scenario and correct chunk using the loading screen. That would result in shorter instanciation lag at the end of the loading process, because the world doesn't further need to load data and just has to instanciate all the data.

DasCapschen:
Instead of writing *.save files ourselves, we can simply save the data in tscn files.

Changes:
- instead of writing 1 .save file, write 1 tscn file per chunk
- chunks save THE SAME INFORMATION AS BEFORE, not extra information is added! (especially for multimeshes, those are still generated at runtime, and NOT saved in the file!)
- but: if we move files, change paths, change script variables, etc, Godot will rename everything FOR US, we don't have to manually edit their .save files anymore.
- Empty Chunks are NOT saved
- Editor will automatically convert .save to .tscn's

Comparision for Tutorials:
- save ~ 8kB
- tscn's ~ 12kB

Comparision for Hainfurt:
- save ~833kB
- tscn's ~980kB

*Edit:*
Resolves #150


src/Data/Scripts/TrackObjects.gd
```diff
@@ -40,8 +40,8 @@ func get_data() -> Dictionary:
 	d.attached_rail = attached_rail
 	d.on_rail_position = on_rail_position
 	d.length = length
-	d.objectPath = objectPath
-	d.materialPaths = materialPaths.duplicate()
+	#d.objectPath = objectPath
+	#d.materialPaths = materialPaths.duplicate()
```
HaSa1002:
```suggestion
```

src/Data/Scripts/TrackObjects.gd
```diff
@@ -69,8 +69,8 @@ func set_data(d: Dictionary) -> void:
 	attached_rail = d.attached_rail
 	on_rail_position = d.on_rail_position
 	length = d.length
-	objectPath = d.objectPath
-	materialPaths = d.materialPaths
+	#objectPath = d.objectPath
+	#materialPaths = d.materialPaths
```
HaSa1002:
```suggestion
```

HaSa1002 (changes requested):
minor things

HaSa1002 (approved):
I am in favor of merging this.
The slightly increased file is fine, because we gain much cleaner code from the change. I believe, we will minify the size even more in the future by changing how we build objects.

DasCapschen:
Something still has an old path somewhere and writes res://chunks ...
Also I didn't fix the chunk system in this PR, just re-wrote it to support the new files.
This means all the bugs that are currently in the chunk system are still in this one too.

I now also converted scenarios.cfg to tres files.
- you have `world.tscn` and next to it `world_config.tres`, this is what used to be "world_config" in scenarios.cfg
  - it has an array of all scenarios, release date, author, ...
- a new folder `scenarios`
  - every scenario gets its own `scenario name.tres` file. It saves scenario name, description, time, trains, signals, etc.

HaSa1002:
Issues noticed in Hainfurt:

- [x] Non existing chunks are tried to be loaded causing error spam (Do dir.file_exists):
	```
	Loading resource: user://editor/hainfurt/Worlds/Hainfurt/chunks/chunk_8_1.tscn
	ERROR: Cannot open file 'user://editor/hainfurt/Worlds/Hainfurt/chunks/chunk_8_1.tscn'.
   	At: scene/resources/resource_format_text.cpp:1232
	ERROR: Failed loading resource: user://editor/hainfurt/Worlds/Hainfurt/chunks/chunk_8_1.tscn. Make sure resources have been 		imported by opening the project in the editor at least once.
   	At: core/io/resource_loader.cpp:282
	```
- [x] Notes for authors was lost
- [x] Scenarios with `:` weren't saved (Convert those to `_` maybe and do an `assert(file_name.is_valid_filename)`?)
- [ ] `Hainfurt.tscn` may include a bunch of unneeded adds (though this may be correct and not caused by this PR. Please check https://github.com/Libre-TrainSim/hainfurt/tree/tscn-chunks)

Positive effects:
 * World loading and chunk spawning is faster.
 * We can, in the future, easily load chunks on world load
 * We don't need to maintain serialisation code

DasCapschen:
- You are right about the chunks, I'll add a check for the file.
- ah, notes are in the .save file... well that's inconvenient... 
- is there a function to sanitize a string to a valid path? I already replace / and \, guess I'll add :, what other symbols do we need to strip?
- What do you mean with unneeded adds?

DasCapschen:
Oh I see now. Looks like it added all the script variables to the tscn file, even if they are the default value. Maybe this is a Godot 3.4 thing? :thinking:

HaSa1002:
> * is there a function to sanitize a string to a valid path? I already replace / and , guess I'll add :, what other symbols do we need to strip?

Nope. Maybe this can help: https://docs.godotengine.org/en/stable/classes/class_string.html?highlight=String#class-string-method-is-valid-filename

> * What do you mean with unneeded adds?

See https://github.com/Libre-TrainSim/hainfurt/commit/b6a54d0743fed0959bfbbeee62291e926be6ef4f#diff-54455ae5b354136ade47699cbb43ac891341d4ba8beb4ffc2a539b2b06da3ee0

HaSa1002:
> Oh I see now. Looks like it added all the script variables to the tscn file, even if they are the default value. Maybe this is a Godot 3.4 thing? 🤔

I did my tests in 3.3.4

DasCapschen:
notes, file_exists and invalid filename should now be fixed.

don't know why godot adds a bunch of stuff to the tscn

Jean28518:
Concept is okay for me!

DasCapschen:
Should be ready now. Please test both ingame and editor! Also try converting an old map and exporting.

World Origin Shift is kind of slow, but that's because *all* Rails and *all* TrackObjects update their MultiMesh...
For some reason, simply changing the position of those does not work...
I will investigate this in the future, it's good enough for now. v0.10 can improve on it.

Chunks are now added into the tree as a single node, like so:
This is potentially faster than adding many nodes individually. It also makes managing chunks a lot easier.
```
World
|--- Chunks
      |--- chunk_0_0
      |      \--- TrackObjects
      |--- chunk_1_0
             \--- TrackObjects
```

src/Data/Scripts/Resources/track_scenario.gd
```diff
@@ -0,0 +1,24 @@
...
+extends Resource
+
+# String -> int
+# I don't like it, but this is what OS.get_time() would return
+# so let's stick to Godots chosen way of doing it.
+export(Dictionary) var time := {
+	"hour": 0,    # TimeH
+	"minute": 0,  # TimeM
+	"second": 0   # TimeS
+}
```
HaSa1002: `OS.get_system_time_secs`?
Thus use int as seconds. We could modulo just fine

src/Data/Scripts/Resources/track_scenario.gd
```diff
@@ -0,0 +1,24 @@
...
+export(String) var title: String = ""
+export(String) var description: String = ""
+export(float) var duration: int = 0  # Minutes?
```
HaSa1002:
```suggestion
export var duration: int = 0  # seconds
```
Goes hand in hand with comments two above

src/Data/Scripts/Resources/track_scenario.gd
```diff
@@ -0,0 +1,24 @@
+class_name TrackScenario
+extends Resource
+
+# String -> int
+# I don't like it, but this is what OS.get_time() would return
+# so let's stick to Godots chosen way of doing it.
+export(Dictionary) var time := {
```
HaSa1002:
```suggestion
export var time := {
```

src/Data/Scripts/Resources/track_scenario.gd
```diff
@@ -0,0 +1,24 @@
...
+	"hour": 0,    # TimeH
+	"minute": 0,  # TimeM
+	"second": 0   # TimeS
+}
+export(String) var title: String = ""
+export(String) var description: String = ""
```
HaSa1002:
```suggestion
export var title := ""
export var description : = ""
```
Export recognizes type. Use implicit types where types are evident.

src/Data/Scripts/Resources/track_scenario.gd
```diff
@@ -0,0 +1,24 @@
...
+export(float) var duration: int = 0  # Minutes?
+
+export(float) var train_length: float = 0.0  # why???
```
HaSa1002: ¯\\_(ツ)_/¯

```suggestion
export var train_length: float = 0.0  # why???
```

src/Data/Scripts/Resources/track_scenario.gd
```diff
@@ -0,0 +1,24 @@
...
+
+# Dict[String, Dict[String, Any]]
+# see World.gd spawn_train()
+export(Dictionary) var trains := {}
+
+# Dict[String, Dict[String, Any]]
+# see Signal.gd get_scenario_data()
+export(Dictionary) var signals := {}
```
HaSa1002:
```suggestion
export var trains := {}

# Dict[String, Dict[String, Any]]
# see Signal.gd get_scenario_data()
export var signals := {}
```

src/Data/Scripts/Resources/world_config.gd
```diff
@@ -0,0 +1,21 @@
...
+extends Resource
+
+
+# I don't like it, but this is what OS.get_date() would return
+# so let's stick to Godots chosen way of doing it.
+export(Dictionary) var release_date := {
+	"year": 2003,
+	"month": 12,
+	"day": 8
+}
```
HaSa1002: see comment in track_scenario.
```suggestion
export release_date: int = 1637869391
```

src/Data/Scripts/Resources/world_config.gd
```diff
@@ -0,0 +1,21 @@
...
+export(String) var author := "Sido"
+export(String) var track_description := "My world, my track, my train, my scenario; from the 1st to the 16th stop."
```
HaSa1002: lol
```suggestion
export var author := "Sido"
export var track_description := "My world, my track, my train, my scenario; from the 1st to the 16th stop."
```

src/Data/Scripts/Resources/world_config.gd
```diff
@@ -0,0 +1,21 @@
...
+
+export(String, FILE, "*.png,*.jpeg,*.jpg,*.bmp,*.gif") var thumbnail_path := ""
+
+# path to the individual scenario.tres's
+export(Array, String, FILE, "*.tres,*.res") var scenarios := []
+
+# shown in the editor
+export(String) var notes := ""
```
HaSa1002:
```suggestion
export var notes := ""
```

src/Data/Scripts/TrackObjects.gd
```diff
@@ -110,37 +106,31 @@ func _exit_tree() -> void:
...
-	update()
+func make_mesh_unique():
+	if multimesh.mesh == null and objectPath != "":
+		multimesh.mesh = load(objectPath)
+
+	if multimesh.mesh != null:
+		multimesh.mesh = multimesh.mesh.duplicate()  # yes, this is necessary!
+		var count: int = int(min(multimesh.mesh.get_surface_count(), materials.size()))
+		for i in range(count):
+			multimesh.mesh.surface_set_material(i, materials[i])
```
HaSa1002: Use the material override of the node.
https://docs.godotengine.org/en/stable/classes/class_geometryinstance.html#class-geometryinstance-property-material-override

src/Data/Scripts/World.gd
```diff
@@ -142,41 +125,44 @@ func get_signal_scenario_data() -> Dictionary:
...
+	timeMinute = current_scenario.time["minute"]
+	timeSecond = current_scenario.time["second"]
 	time = [timeHour,timeMinute,timeSecond]
 
-	apply_scenario_to_signals(scenario["Signals"])
+	apply_scenario_to_signals(current_scenario.signals)
 
 	## SPAWN TRAINS:
-	for train in scenario["Trains"].keys():
+	for train in current_scenario.trains.keys():
```
HaSa1002:
```suggestion
	for train in current_scenario.trains:
```
No need for keys. Train will be a key nontheless

src/Data/Scripts/World.gd
```diff
@@ -142,41 +125,44 @@ func get_signal_scenario_data() -> Dictionary:
...
 		spawnTrain(train)
 
-	jEssentials.call_delayed(1, $Players/Player, "show_textbox_message", [TranslationServer.translate(scenario["Description"])])
-#	$Players/Player.show_textbox_message(TranslationServer.translate(scenario["Description"]))
+	jEssentials.call_delayed(1, $Players/Player, "show_textbox_message", [TranslationServer.translate(current_scenario.description)])
```
HaSa1002:
```suggestion
	jEssentials.call_delayed(1, $Players/Player, "show_textbox_message", [tr(current_scenario.description)])
```
FYI, I don't like this way of calling functions

src/Data/Scripts/World.gd
```diff
@@ -142,41 +125,44 @@ func get_signal_scenario_data() -> Dictionary:
...
 	if $Players.has_node(trainName):
 		Logger.err("Train is already loaded! - Aborted loading...", trainName)
 		return
-	var sData: Dictionary = $jSaveModuleScenarios.get_value("scenario_data")
-	var scenario: Dictionary = sData[currentScenario]
-	var spawnTime: Array = scenario["Trains"][trainName]["SpawnTime"]
-	if scenario["Trains"][trainName]["SpawnTime"][0] != -1 and not (spawnTime[0] == time[0] and spawnTime[1] == time[1] and spawnTime[2] == time[2]):
-		Logger.log("Spawn Time of "+trainName + " not reached, spawning later...")
+
+	var train = current_scenario.trains[trainName]
```
HaSa1002:
```suggestion
	var train: Dictionary = current_scenario.trains[trainName]
```

src/Data/Scripts/World.gd
```diff
@@ -195,23 +181,23 @@ func spawnTrain(trainName: String) -> void:
...
 		new_player.stations["passed"].append(false)
-	new_player.despawnRail = scenario["Trains"][trainName]["DespawnRail"]
+	new_player.despawnRail = train["DespawnRail"]
 	new_player.ai = trainName != "Player"
-	new_player.initialSpeed = Math.kmHToSpeed(scenario["Trains"][trainName].get("InitialSpeed", 0))
-	if scenario["Trains"][trainName].get("InitialSpeedLimit", -1) != -1:
-		new_player.currentSpeedLimit = scenario["Trains"][trainName].get("InitialSpeedLimit", -1)
+	new_player.initialSpeed = Math.kmHToSpeed(train.get("InitialSpeed", 0))
+	if train.get("InitialSpeedLimit", -1) != -1:
+		new_player.currentSpeedLimit = train.get("InitialSpeedLimit", -1)
```
HaSa1002: What is the default speed limit?
Do we really need the check?

src/Data/Scripts/chunk.gd
```diff
@@ -0,0 +1,43 @@
+extends Spatial
+
+export (bool) var is_empty := true
+export (bool) var generate_grass := true  # set this to false when we have terrain
+
+export (Vector3) var chunk_position: Vector3
```
HaSa1002:
```suggestion
export var is_empty := true
export var generate_grass := true  # set this to false when we have terrain

export var chunk_position: Vector3
```

src/Data/Scripts/chunk.gd
```diff
@@ -0,0 +1,43 @@
...
+export (Array, String) var rails := []
+
+
+func _ready() -> void:
+	translation = Vector3(0, 0, 0)
+	if not generate_grass:
+		$DefaultGrass.queue_free()
+	else:
+		$DefaultGrass.translation = chunk_position * 1000  # 1000 = ChunkManager.chunk_size
+		$DefaultGrass.translation.y = -0.5
```
HaSa1002, line 19: isn't the grass a child of the the chunk?

src/Data/Scripts/chunk.gd
```diff
@@ -0,0 +1,43 @@
...
+export (bool) var generate_grass := true  # set this to false when we have terrain
+
+export (Vector3) var chunk_position: Vector3
+
+# array of node names
+export (Array, String) var rails := []
+
+
+func _ready() -> void:
+	translation = Vector3(0, 0, 0)
```
HaSa1002, line 14: why don't we set the translation that gets set below here?

src/Data/Scripts/chunk.gd
```diff
@@ -0,0 +1,43 @@
...
+func _on_world_origin_shifted(delta):
+	if generate_grass:
+		$DefaultGrass.translation += delta
+
+	# FIXME: this does not actually work...
+	#        it would be so nice, if we could just shift the entire chunk node
+	#        instead of all the track objects,
+	#        but I think that will require a TrackObject refactor
+	#        ask me about it for 0.10
+	#translation += delta
```
HaSa1002, line 44:
```suggestion
func _on_world_origin_shifted(delta):
	if generate_grass:
		$DefaultGrass.translation += delta
	update()

	# FIXME: this does not actually work...
	#        it would be so nice, if we could just shift the entire chunk node
	#        instead of all the track objects,
	#        but I think that will require a TrackObject refactor
	#        ask DasCapschen about it for 0.10
	#translation += delta
```

Could be made workable with a change to instance transform? Not a blocker though

src/Data/Scripts/chunk_manager.gd
```diff
@@ -38,8 +32,8 @@ func chunk_to_position(chunk: Vector3) -> Vector3:
 
 
 # needed for serialisation
-func chunk_to_string(chunk: Vector3) -> String:
-	return "%s,%s" % [chunk.x, chunk.z]
+static func chunk_to_string(chunk: Vector3) -> String:
+	return "chunk_%s_%s" % [int(chunk.x), int(chunk.z)]
```
HaSa1002:
```suggestion
	return "chunk_%d_%d" % [chunk.x, chunk.z]
```

src/Data/Scripts/chunk_manager.gd
```diff
@@ -227,14 +222,17 @@ func _finish_chunk_loading():
 
 
 # called deferred from Thread
-func _add_node_to_scene_tree(parent: String, instance: Spatial):
-	assert(world.get_node(parent).get_node_or_null(instance.name) == null)
+func _add_chunk_to_scene_tree(chunk: Spatial):
+	if world.get_node("Chunks").has_node(chunk.name):
+		Logger.err("Chunk already loaded: %s" % chunk.name, self)
+		chunk.free()
+		return
```
HaSa1002: Bug IMO

src/Data/Scripts/chunk_manager.gd
```diff
@@ -263,25 +261,20 @@ func _unload_old_chunks():
...
-			if child.get_meta("chunk_pos") in chunks_to_unload:
-				child.free()
-
 	# remove unloaded chunks
-	for chunk in chunks_to_unload:
-		loaded_chunks.erase(chunk)
+	for chunk_pos in chunks_to_unload:
+		var chunk = world.get_node("Chunks").get_node_or_null(chunk_to_string(chunk_pos))
+		if is_instance_valid(chunk):
+			chunk.free()
```
HaSa1002:
```suggestion
			chunk.queue_free()
```
Why not queue_free? Aren't they still in tree?

src/Data/Scripts/chunk_manager.gd
```diff
@@ -263,25 +261,20 @@ func _unload_old_chunks():
...
+		loaded_chunks.erase(chunk_pos)
+
 	_chunk_mutex.unlock()
 
 
 # handle loading chunks from disk on a separate thread to avoid lags
 func _chunk_loader_thread(_void):
+	var dir = Directory.new()
```
HaSa1002, line 279:
```suggestion
	var dir = Directory.new()
	if dir.open("res://") != OK:
		Logger.err("Cannot open resource directory.", self)
		return
```
Not opening directories is undefined behaviour.

src/Data/Scripts/chunk_manager.gd
```diff
@@ -294,60 +287,24 @@ func _chunk_loader_thread(_void):
...
 				continue
-			var chunk: Dictionary = _jsavemodule.get_value(chunk_to_string(chunk_pos), {"empty": true})
 
-			_generate_landscape(chunk, chunk_pos)
+			var chunk_file
+			if Root.Editor:
+				chunk_file = find_parent("Editor").current_track_path.get_base_dir().plus_file("chunks")
+			else:
+				chunk_file = Root.currentTrack.get_base_dir().plus_file("chunks")
+			chunk_file = chunk_file.plus_file(chunk_to_string(chunk_pos)) + ".tscn"
```
HaSa1002: this is not thread safe
```suggestion
			var chunk_file := ""
			if Root.Editor:
				chunk_file = find_parent("Editor").current_track_path.get_base_dir().plus_file("chunks")
			else:
				chunk_file = Root.currentTrack.get_base_dir().plus_file("chunks")
			chunk_file = chunk_file.plus_file(chunk_to_string(chunk_pos)) + ".tscn"
```

src/Data/Scripts/chunk_manager.gd
```diff
@@ -294,60 +287,24 @@ func _chunk_loader_thread(_void):
...
 
-			if chunk.has("empty"):
-				continue
+			var chunk
```
HaSa1002:
```suggestion
			var chunk: Chunk
```
If Chunk has had a type hint.

src/Data/UI/play_menu.gd
```diff
@@ -44,49 +44,49 @@ func _on_Tracks_item_selected(index: int) -> void:
...
 func _on_Scenarios_item_selected(index: int) -> void:
 	currentScenario = $Play/Selection/Scenarios/Scenarios.get_item_text(index)
-	var sData: Dictionary = $jSaveModule.get_value("scenario_data")
-	$Play/Info/Description.text = tr(sData[currentScenario]["Description"])
-	$Play/Info/Info/Duration.text = "%s: %s min" % [tr("MENU_DURATION"), sData[currentScenario]["Duration"]]
+
+	var scenario = load(world_config.scenarios[index])
+
+	$Play/Info/Description.text = tr(scenario.description)
+	$Play/Info/Info/Duration.text = "%s: %s min" % [tr("MENU_DURATION"), scenario.duration]
```
HaSa1002:
```suggestion
	$Play/Info/Info/Duration.text = "%s: %d min" % [tr("MENU_DURATION"), scenario.duration]
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -1,111 +1,172 @@
...
+
+var world_config_path: String
+var world_config: WorldConfig
+var scenarios: Dictionary
 
 
 func _ready() -> void:
-	save_path = find_parent("Editor").current_track_path + "-scenarios.cfg"
-	$jSaveModule.set_save_path(save_path)
+	var dir = Directory.new()
```
HaSa1002, line 14: `dir.open`

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -1,111 +1,172 @@
...
+	current_scenario = sName
 	update_scenario_list()
 	Logger.log("Scenario added.")
 
 
 func _on_RenameScenario_pressed() -> void:
 	var sName: String = $Scenarios/VBoxContainer/HBoxContainer/LineEdit.text
-	if currentScenario == "" or sName == "" or check_duplicate_scenario(sName) or sName == currentScenario:
+	if current_scenario == "" or sName == "" or check_duplicate_scenario(sName) or sName == current_scenario:
+		Logger.err("Scenario %s cannot be renamed to %s!" % [current_scenario, sName], self)
```
HaSa1002:
```suggestion
		Logger.err("Scenario '%s' cannot be renamed to '%s'!" % [current_scenario, sName], self)
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -1,111 +1,172 @@
...
-	var sData: Dictionary = $jSaveModule.get_value("scenario_data", {})
-	sData[sName] = sData[currentScenario]
-	$jSaveModule.save_value("scenario_data", sData)
-	$jSaveModule.save_value("scenario_list", scenarioList)
-	$jSaveModule.write_to_disk()
-	currentScenario = sName
+
+	var scenario_config = scenarios[current_scenario]
+
+	var dir := Directory.new()
```
HaSa1002, line 96: `dir.open`

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -1,111 +1,172 @@
...
+	scenarios.erase(current_scenario)
+
+	if dir.file_exists(old_scenario_path):
+		dir.remove(old_scenario_path)
+
+	world_config.scenarios.erase(old_scenario_path)
+	world_config.scenarios.append(new_scenario_path)
+	save_world_config()
+
+	Logger.log("Scenario %s successfully renamed to %s." % [current_scenario, sName])
```
HaSa1002:
```suggestion
	Logger.log("Scenario '%s' successfully renamed to '%s'." % [current_scenario, sName])
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -1,111 +1,172 @@
...
+
+	current_scenario = sName
 	update_scenario_list()
 	Logger.log("Scenario added.")
 
 
 func _on_RenameScenario_pressed() -> void:
 	var sName: String = $Scenarios/VBoxContainer/HBoxContainer/LineEdit.text
-	if currentScenario == "" or sName == "" or check_duplicate_scenario(sName) or sName == currentScenario:
+	if current_scenario == "" or sName == "" or check_duplicate_scenario(sName) or sName == current_scenario:
```
HaSa1002:
```suggestion
	if current_scenario.empty() or sName.empty() or check_duplicate_scenario(sName) or sName == current_scenario:
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -1,111 +1,172 @@
...
+	world_config.scenarios.append(scenario_path)
+	save_world_config()
+
+	current_scenario = sName
 	update_scenario_list()
 	Logger.log("Scenario added.")
 
 
 func _on_RenameScenario_pressed() -> void:
 	var sName: String = $Scenarios/VBoxContainer/HBoxContainer/LineEdit.text
```
HaSa1002, line 89:
```suggestion
	var sName: String = $Scenarios/VBoxContainer/HBoxContainer/LineEdit.text.strip_edges()
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -1,111 +1,172 @@
...
+	save_world_config()
+
+	Logger.log("Scenario %s duplicated." % current_scenario)
+	current_scenario = sName
 	update_scenario_list()
 
 
 func _on_DeleteScenario_pressed() -> void:
-	if currentScenario == "":
+	if current_scenario == "":
```
HaSa1002:
```suggestion
	if current_scenario.empty():
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -1,111 +1,172 @@
...
-	$jSaveModule.write_to_disk()
-	currentScenario = ""
+
+	scenarios.erase(current_scenario)
+
+	var scenario_path: String = get_scenario_path(current_scenario)
+	world_config.scenarios.erase(scenario_path)
+	save_world_config()
+
+	var dir = Directory.new()
```
HaSa1002, line 151: `dir.open`

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -114,51 +175,52 @@ func _process(delta: float) -> void:
...
-	$Scenarios/VBoxContainer/ResetSignals.visible = currentScenario != ""
+	$Scenarios/VBoxContainer/Settings.visible = current_scenario != ""
+	$Scenarios/VBoxContainer/Label2.visible = current_scenario != ""
+	$Scenarios/VBoxContainer/SaveSignalData.visible = current_scenario != ""
+	$Scenarios/VBoxContainer/CopySignalDataFrom.visible = current_scenario != ""
+	$Scenarios/VBoxContainer/ResetSignals.visible = current_scenario != ""
 
 
-func get_scenario_settings() -> void: # fills the settings field with saved values
+func update_ui_from_scenario_config() -> void: # fills the settings field with saved values
```
HaSa1002, line 190:
```suggestion
func update_ui_to_scenario_config() -> void: # fills the settings field with saved values
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -114,51 +175,52 @@ func _process(delta: float) -> void:
...
 	clear_general_scenario_settings_fields()
-	var sData: Dictionary = $jSaveModule.get_value("scenario_data", {})
-	if not sData.has(currentScenario):
-		return
-	var s: Dictionary = sData[currentScenario]
 
-	if s.size() == 0:
+	if not scenarios.has(current_scenario):
+		Logger.err("Scenario %s is not in scenario list!" % current_scenario, self)
```
HaSa1002:
```suggestion
		Logger.err("Scenario '%s' is not in scenario list!" % current_scenario, self)
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -114,51 +175,52 @@ func _process(delta: float) -> void:
...
+	$Scenarios/VBoxContainer/Settings/Tab/General/Time/TimeSecond.value = s.time["second"]
+	$Scenarios/VBoxContainer/Settings/Tab/General/TrainLength/SpinBox.value = s.train_length
+	$Scenarios/VBoxContainer/Settings/Tab/General/Description.text = s.description
+	$Scenarios/VBoxContainer/Settings/Tab/General/Duration/SpinBox.value = s.duration
 	Logger.log("Scenario Settings loaded")
 
 
 func save_general_scenario_settings() -> void:
-	if currentScenario == "":
+	if current_scenario == "":
```
HaSa1002:
```suggestion
	if current_scenario.empty():
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -173,20 +235,19 @@ func clear_general_scenario_settings_fields() -> void:
 
 func update_scenario_list() -> void:
 	$Scenarios/VBoxContainer/ItemList.clear()
-	var scenarios: Array = $jSaveModule.get_value("scenario_list", [])
-	for scenario in scenarios:
+	for scenario in scenarios.keys():
```
HaSa1002:
```suggestion
	for scenario in scenarios:
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -173,20 +235,19 @@ func clear_general_scenario_settings_fields() -> void:
...
 
 func update_train_list() -> void:
 	$Scenarios/VBoxContainer/Settings/Tab/Trains/ItemList2.clear()
-	var sData: Dictionary = $jSaveModule.get_value("scenario_data", {})
-	if not sData.has(currentScenario):
-		return
-	if not sData[currentScenario].has("Trains"):
+
+	if not scenarios.has(current_scenario):
+		Logger.err("Scenario %s is not in scenario list!" % current_scenario, self)
```
HaSa1002:
```suggestion
		Logger.err("Scenario '%s' is not in scenario list!" % current_scenario, self)
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -173,20 +235,19 @@ func clear_general_scenario_settings_fields() -> void:
...
 		return
-	var trains: Array = sData[currentScenario]["Trains"].keys()
+
+	var trains: Array = scenarios[current_scenario].trains.keys()
 	for train in trains:
```
HaSa1002:
```suggestion
	for train in scenarios[current_scenario].trains:
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -197,20 +258,20 @@ func _on_SaveGeneral_pressed() -> void:
 
 
 func save_everything() -> void:
-	if currentScenario != "":
+	if current_scenario != "":
```
HaSa1002:
```suggestion
	if !current_scenario.empty():
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -197,20 +258,20 @@ func _on_SaveGeneral_pressed() -> void:
...
 		_on_SaveGeneral_pressed()
-	if currentTrain != "":
+	if current_train != "":
```
HaSa1002:
```suggestion
	if !current_train.empty():
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -197,20 +258,20 @@ func _on_SaveGeneral_pressed() -> void:
...
 func _on_ItemList_item_selected(index: int) -> void:
-	currentScenario = $Scenarios/VBoxContainer/ItemList.get_item_text(index)
-	world.currentScenario = currentScenario
+	current_scenario = $Scenarios/VBoxContainer/ItemList.get_item_text(index)
+	world.current_scenario = current_scenario
 	update_train_list()
-	get_train_settings()
-	get_scenario_settings()
+	load_current_train()
+	update_ui_from_scenario_config()
```
HaSa1002, line 273:
```suggestion
	update_ui_to_scenario_config()
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -219,55 +280,44 @@ func _on_SaveChunks_pressed() -> void:
...
 
-var currentTrain: String = "Player"
 
-func get_train_settings() -> void:
-	var sData: Dictionary = $jSaveModule.get_value("scenario_data", {})
-	if not sData.has(currentScenario):
+var current_train: String = "Player"
+func load_current_train() -> void:
+	if not scenarios.has(current_scenario):
+		Logger.err("Scenario %s is not in scenario list!" % current_scenario, self)
```
HaSa1002:
```suggestion
		Logger.err("Scenario '%s' is not in scenario list!" % current_scenario, self)
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -282,10 +332,13 @@ func get_train_settings() -> void:
 	$Scenarios/VBoxContainer/Settings/Tab/Trains/GridContainer/DespawnRail.text = train["DespawnRail"]
 	$Scenarios/VBoxContainer/Settings/Tab/Trains/GridContainer/InitialSpeed.value = train.get("InitialSpeed", 0)
 	$Scenarios/VBoxContainer/Settings/Tab/Trains/GridContainer/InitialSpeedLimit.value = train.get("InitialSpeedLimit", -1)
-	Logger.log("Train "+ currentTrain + " loaded.")
+	Logger.log("Train "+ current_train + " loaded.")
+
 
+func save_current_train() -> void:
+	if current_scenario == "":
```
HaSa1002:
```suggestion
	if current_scenario.empty():
```
Honestly a lot of those checks are probably bug hiding

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -428,61 +371,58 @@ func _on_NewTrain_pressed() -> void:
...
-			return
-	get_train_settings()
-	currentTrain = trainName
-	set_train_settings()
-	## Delete "Old Train"
-	delete_train(oldTrain)
-	update_train_list()
+
+	if scenarios[current_scenario].trains.has(trainName):
+		Logger.err("Cannot rename train %s to %s, already exists." % [oldTrain, trainName], self)
```
HaSa1002:
```suggestion
		Logger.err("Cannot rename train '%s' to '%s', already exists." % [oldTrain, trainName], self)
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -428,61 +371,58 @@ func _on_NewTrain_pressed() -> void:
...
+		return
+
+	load_current_train()      # load old train
+	current_train = trainName  # rename
+	save_current_train()      # save new train
+	delete_train(oldTrain)    # delete old train
+	update_train_list()       # update ui
```
HaSa1002:
```suggestion
	load_current_train() # load old train
	current_train = trainName # rename
	save_current_train() # save new train
	delete_train(oldTrain) # delete old train
	update_train_list() # update ui
```
Please no.

HaSa1002: or kill the comments completely. The methods are kinda explicit already

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -428,61 +371,58 @@ func _on_NewTrain_pressed() -> void:
...
+	current_train = current_train + " (Duplicate)"
+	$Scenarios/VBoxContainer/Settings/Tab/Trains/ItemList2.add_item(current_train)
+	save_current_train()
 
 
 func delete_train(train: String) -> void:
-	var sData: Dictionary = $jSaveModule.get_value("scenario_data", {})
-	if not sData.has(currentScenario):
+	if not scenarios.has(current_scenario):
+		Logger.err("Scenario %s is not in scenario list!" % current_scenario, self)
```
HaSa1002:
```suggestion
		Logger.err("Scenario '%s' is not in scenario list!" % current_scenario, self)
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -428,61 +371,58 @@ func _on_NewTrain_pressed() -> void:
...
 		return
-	if not sData[currentScenario].has("Trains"):
+	if not scenarios[current_scenario].trains.has(train):
+		Logger.warn("Cannot delete Train " + train + ", does not exist...", self)
```
HaSa1002:
```suggestion
		Logger.warn("Cannot delete non existent train `%s`." % train, self)
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -533,47 +472,39 @@ func _on_SaveSignalData_pressed() -> void:
...
 func save_signal_data_to_current_scenario() -> void:
 	var signal_data: Dictionary = world.get_signal_scenario_data()
-	var sData: Dictionary = $jSaveModule.get_value("scenario_data", {})
-	if not sData.has(currentScenario):
-		sData[currentScenario] = {}
-	sData[currentScenario]["Signals"] = signal_data
-	$jSaveModule.save_value("scenario_data", sData)
-	$jSaveModule.write_to_disk()
+	if not scenarios.has(current_scenario):
+		Logger.err("Scenario %s is not in scenario list!" % current_scenario, self)
```
HaSa1002:
```suggestion
		Logger.err("Scenario '%s' is not in scenario list!" % current_scenario, self)
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -533,47 +472,39 @@ func _on_SaveSignalData_pressed() -> void:
...
 
 
 func _on_CopyAndOverwriteSignalDataFrom_pressed() -> void:
-	var scenario_list: Array = $jSaveModule.get_value("scenario_list", [])
-	if scenario_list.size() < 2:
+	if scenarios.size() < 2:
 		return
 	$Scenarios/VBoxContainer/CopySignalDataFrom/PopupMenu.clear()
-	for scenario in scenario_list:
+	for scenario in scenarios.keys():
```
HaSa1002:
```suggestion
	for scenario in scenarios:
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -533,47 +472,39 @@ func _on_SaveSignalData_pressed() -> void:
...
+		Logger.err("Scenario %s is not in scenario list!" % current_scenario, self)
+		return
+	scenarios[current_scenario].signals = signal_data
+	save_current_scenario()
 
 
 func _on_CopyAndOverwriteSignalDataFrom_pressed() -> void:
-	var scenario_list: Array = $jSaveModule.get_value("scenario_list", [])
-	if scenario_list.size() < 2:
+	if scenarios.size() < 2:
```
HaSa1002, line 482: warn?

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -533,47 +472,39 @@ func _on_SaveSignalData_pressed() -> void:
...
 		$Scenarios/VBoxContainer/CopySignalDataFrom/PopupMenu.add_item(scenario)
 	$Scenarios/VBoxContainer/CopySignalDataFrom/PopupMenu.show()
 
 
 func load_signal_data_from_current_scenario_to_world() -> void:
-	if currentScenario == "":
-		return
-	var scenario_data: Dictionary = $jSaveModule.get_value("scenario_data", {})
-	if not scenario_data.has(currentScenario):
+	if current_scenario == "":
```
HaSa1002:
```suggestion
	if current_scenario.empty():
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -533,47 +472,39 @@ func _on_SaveSignalData_pressed() -> void:
...
-	var signal_data: Dictionary = current_scenario_data["Signals"]
-	world.apply_scenario_to_signals(signal_data)
+	world.apply_scenario_to_signals(scenarios[current_scenario].signals)
 
 
 func _on_PopupMenu_Copy_SignalDataFrom_index_pressed(index: int) -> void:
 	var scenario_source: String = $Scenarios/VBoxContainer/CopySignalDataFrom/PopupMenu.get_item_text(index)
-	if scenario_source == currentScenario:
+	if scenario_source == current_scenario:
 		jEssentials.show_message("Nothing done. Can't copy and overwrite from current scenario to current scenario.")
```
HaSa1002, line 504: Don't we have a show_message here. And we miss a log, probably

src/Editor/Docks/RailAttachments/RailAttachments.gd
```diff
@@ -288,18 +288,20 @@ func update_material_list() -> void:
 	if not is_instance_valid(mesh_instance) or mesh_instance.mesh.resource_path != currentTO.objectPath:
 		mesh_instance = MeshInstance.new()
 		mesh_instance.mesh = load(currentTO.objectPath)
-	var material_array: Array = currentTO.materialPaths
+	var material_array: Array = currentTO.materials
 	for i in range(mesh_instance.get_surface_material_count()):
-		if i < material_array.size() and ResourceLoader.exists(material_array[i]):
-			mesh_instance.set_surface_material(i, load(material_array[i]))
+		if i < material_array.size():
+			mesh_instance.set_surface_material(i, material_array[i])
```
HaSa1002, line 294: Do using material overwrite in the node?

src/Editor/Scripts/Editor.gd
```diff
@@ -15,14 +15,167 @@ var selected_object_type: String = ""
...
 	if !load_world():
 		return
 
+	_port_to_new_scenario_resource()
+	_port_to_new_chunk_system()
+
+
+func _port_to_new_chunk_system() -> void:
+	var save_file = current_track_path + ".save"
+	var dir = Directory.new()
```
HaSa1002, line 31: `dir.open`

src/Editor/Scripts/Editor.gd
```diff
@@ -15,14 +15,167 @@ var selected_object_type: String = ""
...
+
+			if to_instance.multimesh.mesh == null:
+				Logger.err("Could not load object %s" % track_obj.data.objectPath, self)
+
+			if to_instance.multimesh.mesh != null:
+				var i = 0
+				for material_path in track_obj.data.materialPaths:
+					to_instance.materials.append(load(material_path))
+					to_instance.multimesh.mesh.surface_set_material(i, load(material_path))
+					i += 1
```
HaSa1002:
```suggestion
			if to_instance.multimesh.mesh != null:
				var i = 0
				for material_path in track_obj.data.materialPaths:
					to_instance.materials.append(load(material_path))
					to_instance.multimesh.mesh.surface_set_material(i, load(material_path))
					i += 1
			else:
				Logger.err("Could not load object '%s'" % track_obj.data.objectPath, self)
```

src/Editor/Scripts/Editor.gd
```diff
@@ -15,14 +15,167 @@ var selected_object_type: String = ""
...
+			to_instance.set_data(track_obj.data)
+
+			to_instance.transform = track_obj.transform
+
+			new_chunk.get_node("TrackObjects").add_child(to_instance)
+			to_instance.owner = new_chunk
+
+		if len(new_chunk.rails) > 0 \
+		or new_chunk.get_node("Buildings").get_child_count() > 0 \
+		or new_chunk.get_node("TrackObjects").get_child_count() > 0:
```
HaSa1002:
```suggestion
				or new_chunk.get_node("Buildings").get_child_count() > 0 \
				or new_chunk.get_node("TrackObjects").get_child_count() > 0:
```
https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_styleguide.html#formatting

src/Editor/Scripts/Editor.gd
```diff
@@ -15,14 +15,167 @@ var selected_object_type: String = ""
...
+			new_chunk.is_empty = false
+			var packed_chunk = PackedScene.new()
+			packed_chunk.pack(new_chunk)
+			var path = current_track_path.get_base_dir().plus_file("chunks").plus_file(ChunkManager.chunk_to_string(old_chunk.position)) + ".tscn"
```
HaSa1002:
```suggestion
			var packed_chunk := PackedScene.new()
			packed_chunk.pack(new_chunk)
			var path: String = current_track_path.get_base_dir().plus_file("chunks").plus_file(ChunkManager.chunk_to_string(old_chunk.position)) + ".tscn"
```

src/Editor/Scripts/Editor.gd
```diff
@@ -15,14 +15,167 @@ var selected_object_type: String = ""
...
+
+		dir.remove(save_file)
+		new_chunk.queue_free()
+
+	send_message("Chunks were ported to v0.9, please close and reload the track.")
+
+
+func _port_to_new_scenario_resource() -> void:
+	var scenario_file := current_track_path + "-scenarios.cfg"
+	var dir = Directory.new()
```
HaSa1002, line 115: `dir.open`

src/Editor/Scripts/Editor.gd
```diff
@@ -15,14 +15,167 @@ var selected_object_type: String = ""
...
+
+	var jsavemodule = Node.new()
+	jsavemodule.set_script(load("res://addons/jean28518.jTools/jSaveManager/jSaveModule.gd"))
+	jsavemodule.set_save_path(scenario_file)
+
+	var world_config_path = current_track_path + "_config.tres"
+	var world_config = WorldConfig.new()
+
+	var old_world_config = jsavemodule.get_value("world_config")
+	var old_release_date = old_world_config.get("ReleaseDate", [1,1,2021])
```
HaSa1002:
```suggestion
	var world_config_path: String = current_track_path + "_config.tres"
	var world_config := WorldConfig.new()

	var old_world_config: Dictionary = jsavemodule.get_value("world_config")
	var old_release_date: Array = old_world_config.get("ReleaseDate", [1,1,2021])
```

src/Editor/Scripts/Editor.gd
```diff
@@ -15,14 +15,167 @@ var selected_object_type: String = ""
...
+	world_config.author = old_world_config.get("Author", "")
+	world_config.track_description = old_world_config.get("TrackDesciption", "")
+	world_config.scenarios = []
+
+	var scenarios = jsavemodule.get_value("scenario_data")
+	for key in scenarios.keys():
+		if key == "":  # there is always an empty scenario, pls ignore it...
+			continue
+		var old_scenario = scenarios[key]
+		var new_scenario = TrackScenario.new()
```
HaSa1002:
```suggestion
	var scenarios: Dictionary = jsavemodule.get_value("scenario_data")
	for key in scenarios:
		if key.empty()":  # there is always an empty scenario, pls ignore it...
			continue
		var old_scenario: Dictionary = scenarios[key]
		var new_scenario := TrackScenario.new()
```
What's the type of `old_scenario`?

src/Editor/Scripts/Editor.gd
```diff
@@ -15,14 +15,167 @@ var selected_object_type: String = ""
...
+		new_scenario.time["second"] = old_scenario.get("TimeS", 0)
+		new_scenario.signals = old_scenario.get("Signals", {})
+		new_scenario.trains = old_scenario.get("Trains", {})
+		new_scenario.train_length = old_scenario.get("TrainLength", 0)
+
+		key = key.replace("/", "_").replace("\\", "_").replace(":", "_")\
+				 .replace("?", "_").replace("*", "_").replace("\"", "_")\
+				 .replace("|", "_").replace("%", "_").replace("<", "_")\
+				 .replace(">", "_")
+		var scenario_path = scenario_dir.plus_file("%s.tres" % key)
```
HaSa1002:
```suggestion
		var scenario_path: String = scenario_dir.plus_file("%s.tres" % key)
```

src/Editor/Scripts/Editor.gd
```diff
@@ -15,14 +15,167 @@ var selected_object_type: String = ""
...
+		if ResourceSaver.save(scenario_path, new_scenario) != OK:
+			Logger.err("Could not save scenario at %s!" % scenario_path, self)
+			continue
+
+		world_config.scenarios.append(scenario_path)
+
+	var save_file = current_track_path + ".save"
```
HaSa1002:
```suggestion
	var save_file: String = current_track_path + ".save"
```

src/Editor/Scripts/editor_configuration.gd
```diff
@@ -43,15 +43,22 @@ func _find_content():
...
+	dir.make_dir_recursive(mod_path.plus_file(worlds_path).plus_file("scenarios"))
 	dir.copy("res://Data/Modules/World-Pattern.tscn", \
-			"%s%s/%s.tscn" % [mod_path, worlds_path, entry_name])
-	dir.copy("res://Data/Modules/World-Pattern.save", \
-			"%s%s/%s.save" % [mod_path, worlds_path, entry_name])
+			"%s.tscn" % mod_path.plus_file(worlds_path).plus_file(entry_name))
+
+	var chunk_0_0 = preload("res://Data/Modules/chunk_prefab.tscn").instance()
+	chunk_0_0.rails = ["Rail"]
+	var packed_chunk = PackedScene.new()
```
HaSa1002:
```suggestion
	var chunk_0_0 := preload("res://Data/Modules/chunk_prefab.tscn").instance() as Spatial
	chunk_0_0.rails = ["Rail"]
	var packed_chunk := PackedScene.new()
```
as `what`?

src/Worlds/German-Safety-Tutorials/SafetyTutorialScript.gd
```diff
@@ -15,6 +15,9 @@ var pzb_module: Node
 
 var init_done: bool = false
 func init() -> void:
+	if Root.Editor:
+		return
```
HaSa1002, line 19: What about ready or yield in ready?
We could do `set_process(!Root.Editor)` there as well.

src/Worlds/Signal-Tutorial/Scripting.gd
```diff
@@ -8,6 +8,10 @@ var message_sent: bool = false
 
 
 func _process(delta: float) -> void:
+	if Root.Editor:
+		set_process(false)
+		return
```
HaSa1002, line 13: _ready?

HaSa1002 (changes requested):
Mostly style.
Though I still have to do some tests

HaSa1002:
Game and editor as well as conversion still work. Well done
grafik

HaSa1002:
Oh wait. The generated multi meshes still get saved

src/Data/Scripts/Resources/track_scenario.gd
DasCapschen: IIRC, Duration is entered by the user and is AFAIK only so players know roughly what to expect when starting a scenario.
So it should probably be minutes

src/Data/Scripts/Resources/track_scenario.gd
DasCapschen: well, the world script does everything using an [Hour, Minute, Second] Array for tracking time... 😅

src/Data/Scripts/TrackObjects.gd
DasCapschen: Material Override does not have seperate surfaces though. It overrides everything with 1 Material.
We need multiple Materials per Object though.

src/Data/Scripts/chunk.gd
DasCapschen: Yes, Grass is a child of the chunk.
But objects inside the chunk (with the exception of grass) are not stored relative to the chunk, but rather in global coordinates.
This means the chunk itself needs to be at (0,0,0).
But Grass *is* relative to the chunk, so we need to move it.

src/Data/Scripts/chunk.gd
DasCapschen: I tried moving all TrackObject instances, but that does not work. 
It only works when I trigger update() on all Track Objects. I have no idea why, moving transform *should* work.

Also, we should not call update here. Rails already call update on all their assigned TrackObjects.
(and Rails are updated by the signal)

src/Data/Scripts/chunk_manager.gd
DasCapschen: Yes, but it happens some times, usually when there is a huge lag due to world origin transform.

DasCapschen: If I assert this, we get memory leaks at runtime.

src/Data/Scripts/chunk_manager.gd
DasCapschen: free removes them from tree?

src/Data/Scripts/chunk_manager.gd
DasCapschen: We only use "static" function like checking if a file exists though... tbh it is beyond me why those are not static. 
There is no need to open a directory for that...

src/Data/Scripts/chunk_manager.gd
DasCapschen: chunk does not have a class_name

src/Editor/Docks/Configuration/Configuration.gd
DasCapschen: why?

src/Editor/Docks/Configuration/Configuration.gd
DasCapschen: Honestly, checks are just making sure nothing crashes *if* a bug ever occured.

src/Editor/Docks/Configuration/Configuration.gd
DasCapschen: 1) why are you against order? it's good to have them listed below each other... also, 2 spaces before a comment
2) yeah, I guess we can just remove them outright

src/Editor/Docks/Configuration/Configuration.gd
DasCapschen: Maybe? idk, I didn't write this code

src/Editor/Docks/RailAttachments/RailAttachments.gd
DasCapschen: again, material override does not support multiple materials per object

src/Editor/Scripts/Editor.gd
DasCapschen: next time, please link directly to the correct statement, I don't want to search through a huge webpage
https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_styleguide.html#format-multiline-statements-for-readability

src/Editor/Scripts/Editor.gd
DasCapschen: No, sorry, it's not string, it is a dict.

DasCapschen:
> The generated multi meshes still get saved

Well that's bad... that shouldn't have happened... will look into it

src/Worlds/Signal-Tutorial/Scripting.gd
DasCapschen: tbh, all scenario scripts are absolute jank

src/Data/Scripts/chunk_manager.gd
DasCapschen: > When creating a new Directory, its default opened directory will be res://.

https://docs.godotengine.org/en/stable/classes/class_directory.html

src/Editor/Scripts/Editor.gd
DasCapschen: honestly, I find it a lot *less* readable with 2 indentations...
without indentation, you get `if or or` underneath each other, with 2 indents... it's just hell...

src/Data/Scripts/Resources/track_scenario.gd
HaSa1002: ok 0.10 then

src/Data/Scripts/Resources/track_scenario.gd
HaSa1002: ok

src/Data/Scripts/chunk.gd
HaSa1002: 0.10 fix probably

src/Data/Scripts/chunk_manager.gd
HaSa1002: queue_free does that as well, but at the end of the frame

src/Data/Scripts/chunk_manager.gd
HaSa1002: > When creating a new Directory, its default opened directory will be res://. This may change in the future, so it is advised to always use open to initialize your Directory where you want to operate, with explicit error checking.

But ok :) We can still create a ResourceDirectory class and a UserDirectory class

src/Editor/Scripts/Editor.gd
HaSa1002: yeah I agree two indents is not optimal, but the inconsistency is not better as well IMHO. I was confused at first because I parsed them as new statements.

src/Worlds/Signal-Tutorial/Scripting.gd
HaSa1002: true

HaSa1002:
I found these issues:
1. Big Chunk chunk swich breaks the current chunk
2. Not all multimeshes get deleted on convert
3. Let's decrease the big chunk to 5000 again...
4. UI needs to be updated as well

HaSa1002:
Rebase is needed. We can continue with this path as master is open for 1.0 now

PR #237, opened 2021-11-07
Move camera faster the longer you move

### LINKED ISSUE #230: Move editor camera faster the longer you move
As our maps tend to get big, we need to travel fast on some occasions. On the other hand, we do need slow movement to precisely adjust the camera. This can combination is easily achieved when you accelerate the camera over time. It feels natural.

HaSa1002:
Fixes #230


DasCapschen (approved):
lgtm

PR #246, opened 2021-11-12
fix getPathFromTo

### LINKED ISSUE #203: getPathFromToHelper crash
Function `getPathFromToHelper` in Wagon.gd:
```gdscript
if connectionN.is_in_group("PassengerPathNode"):
```
"attempt to call is_in_group on null instance"

DasCapschen:
fixes #203

### DIFF
--- a/src/Data/Modules/PassengerSeat.tscn
+++ b/src/Data/Modules/PassengerSeat.tscn
@@ -14,9 +14,7 @@ size = Vector3( 0.5, 0.5, 0.2 )
 [sub_resource type="SpatialMaterial" id=4]
 albedo_color = Color( 0.00784314, 0.788235, 0, 1 )
 
-[node name="PassengerSeat" type="Spatial" groups=[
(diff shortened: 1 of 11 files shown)


HaSa1002:
IMO those connections should be NodePaths

DasCapschen:
They cannot be NodePaths, because NodePath is always relative to the Node they are saved on, but the path needs to be relative to the Wagon...
Although, I think I can change the code to be relative to the current Node instead of the wagon :thinking:

DasCapschen:
Okay, it is now using NodePaths.

BUT!
I just noticed it is actually still broken.
Passengers never sit down, they only get in the train and then get stuck on some path node.
I mean, it's better than before, where passengers got stuck outside the train, but yeah... :/

HaSa1002 (approved):
I suppose it is better than now. Let's merge it and fix more later :)

HaSa1002 (approved):
works fine

PR #247, opened 2021-11-12
fix scenario to signals

### LINKED ISSUE #244: apply_scenario_to_signals fails because dictionaries are null
World.gd:543:
```gdscript
func apply_scenario_to_signals(signals: Dictionary) -> void:
	## Apply Scenario Data
	for signalN in $Signals.get_children():
		if signals.has(signalN.name):
			signalN.set_scenario_data(signals[signalN.name]) # set_scenario_data wants to have dict, but may get null
```
Workaround, but probably not a fix:
```gdscript
func apply_scenario_to_signals(signals: Dictionary) -> void:
	## Apply Scenario Data
	for signalN in $Signals.get_children():
		if signals.has(signalN.name):
			signalN.set_scenario_data(signals[signalN.name] if signals[signalN.name] != null else {})
```

DasCapschen:
fixes #244

### DIFF
--- a/src/Data/Scripts/World.gd
+++ b/src/Data/Scripts/World.gd
@@ -544,7 +544,7 @@ func apply_scenario_to_signals(signals: Dictionary) -> void:
 	## Apply Scenario Data
 	for signalN in $Signals.get_children():
 		if signals.has(signalN.name):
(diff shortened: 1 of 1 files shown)


HaSa1002 (approved):
What could possibly go wrong 🤷‍♂️

PR #249, opened 2021-11-12
Add very simple scoring system

### LINKED ISSUE #31: Evaluation in the end of the scenario.
That will let you measure, how well you drove the scenario ;)

DasCapschen:
Closes #31 

Scoring works like:
- you get 100 points when you arrive at a station
- you lose `speed - speedlimit` points when you drive too fast
	- example: speed limit = 80, you drive 90, you lose 10 points every 5 seconds until you slow down
- you lose 10 points if you arrive late (less than 1 minute)
- you lose 50 points if you arrive very late (more than 1 minute)
- you lose 50 points if you depart early
- you lose 100 points if you get an emergency braking
- you lose 1000 points if you pass a red light

At the end of the scenario, you get a message telling you the scenario is over and how many points you have scored.
Clicking OK will end the scenario and load the main menu.
grafik
(i used debug mode, so a lot of speeding, that's why it's -800 points 😅 )


src/Data/Scripts/Player.gd
```diff
@@ -156,18 +156,33 @@ var startRail: String # Rail, on which the train is starting. Set by the scenari
...
+const SCORE_ARRIVE_AT_STATION: int = 100  # when you arrive at a station
+# losing points
+const SCORE_MULTIPLIER_TOO_FAST: float = 1.0  # score -= (speed - speedLimit) * multiplier
+const SCORE_PENALTY_RED_LIGHT: int = 1000  # when you drive over a red light
+const SCORE_PENALTY_ARRIVE_LATE: int = 10  # when you arrive at a station late
+const SCORE_PENALTY_ARRIVE_VERY_LATE: int = 50  # when you arrive at a station more than a minute late
+const SCORE_PENALTY_DEPART_EARLY: int = 50  # when you leave a station too early
+const SCORE_PENALTY_EMERGENCY_BRAKE: int = 100  # when you get an emergency braking
+
+### SIGNALS ###
```
HaSa1002, line 174: ouf. Another thing in the player script. Like it isn't big enought already

HaSa1002 (approved):
I am not so sure about the amount of points, but that is easily fixable. Looks good otherwise (except for comment, but sadly has to be this way atm)

src/Data/Scripts/Player.gd
DasCapschen: Yeah I know...
I could make it into its own script running on a node in the player, but then it still needs to be connected to the player by signals. I'm not sure if that makes it any better 🙈 

We can always refactor it later, it's pretty straightforward since it's only 1 variable and it doesn't interact with anything else.

HaSa1002: yeah I know.

src/Data/Scripts/Player.gd
```diff
@@ -156,18 +156,33 @@ var startRail: String # Rail, on which the train is starting. Set by the scenari
...
 signal reverser_changed(reverser_state)
+signal _textbox_closed
 
-func ready() -> void: ## Called by World!
+
+# Called by World!
+func ready() -> void:
 	pause_mode = Node.PAUSE_MODE_PROCESS
 	$Camera.pause_mode = Node.PAUSE_MODE_PROCESS
+
```
```suggestion

```
You reset your editor settings 😅

DasCapschen: Oh, yeah I had to wipe Godot because of some issue, guess I forgot to turn it back on. Will fix.

src/Data/Scripts/Player.gd
```diff
@@ -979,6 +1004,7 @@ var checkSpeedLimitTimer: float = 0
 func checkSpeedLimit(delta: float) -> void:
 	if Math.speedToKmH(speed) > currentSpeedLimit + 5 and checkSpeedLimitTimer > 5:
 		checkSpeedLimitTimer = 0
+		score -= int((Math.speedToKmH(speed) - currentSpeedLimit) * SCORE_MULTIPLIER_TOO_FAST)
```
```suggestion
		score -= int(round((Math.speedToKmH(speed) - currentSpeedLimit) * SCORE_MULTIPLIER_TOO_FAST))

HaSa1002 (changes requested):
Just nitpicking. Works fine otherwise

HaSa1002 (approved):
lgtm

PR #250, opened 2021-11-12
Add difficulty settings

### LINKED ISSUE #132: Add difficulty settings
We should add difficulty settings to the options menu to allow the player to customize their experience. Right now this would probably only involve toggling SiFa on or off. (and I guess we could move simple mode there, instead of at the track selection)
More settings could follow in the future (see #131)

DasCapschen:
Fixes #132 

grafik


HaSa1002 (approved):
lgtm

PR #262, opened 2021-11-21
Refactor Chunk System

### LINKED ISSUE #194: Chunksystem chunk loader is not thread safe
Interactions with the scene tree are not thread safe. Refer to <https://docs.godotengine.org/en/stable/tutorials/threads/thread_safe_apis.html#scene-tree> for more information.

### LINKED ISSUE #233: Grass Tile generation broke
grafik

### LINKED ISSUE #241: Chunks don't work correctly inside Editor

### LINKED ISSUE #258: Wrong grass plane is unloaded
grafik

DasCapschen:
fixes #233, fixes #194, fixes #241, fixes #258 

- separate World.gd and chunk_manager.gd 
  - easier to maintain
  - before: World.gd ~840 lines of code
  - now: World.gd ~330 loc, chunk_manager.gd ~350 loc
- refactored chunk system. It's mostly the same as before, but cleaned up.
  - runs on thread
  - add nodes via `call_deferred`
  - unloading happens synchronously on main thread
  - removed "Flora" code, it was deprecated anyways
  - use `global_transform.origin` instead of `translation` (not 100% sure if this is necessary)
  - signals for: world origin shift, thread finished loading (internal use only), chunks finished loading (shortly after thread finished)

From my testing, it now works properly. I have had no problems on Hainfurt nor in the Editor.
It is still a little laggy when changing big chunks.

- Known bugs:
  - person system broke. people just do not move *at all*. I have no clue why. There are no errors/warnings logged.


src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
+class_name ChunkManager
+extends Node
+
+signal world_origin_shifted(delta)  # emitted when world origin changes
+signal chunks_finished_loading   # emitted when chunks finished loading (loaded_chunks == chunks_to_load)
+
+const world_origin_shift_treshold: int = 5000  # if further than this from origin, recenter world origin
```
HaSa1002:
```suggestion
const world_origin_shift_treshold: int = 131_072 # if further than this from origin, recenter world origin
```
5000 is stupidly low value given this operation is costy and has no real backing. See this table:
grafik
(Taken from this wonderful blog post: https://blog.demofox.org/2017/11/21/floating-point-precision/)

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+const chunk_size: int = 1000  # extend of a chunk in all directions
+
+var world
+var world_origin := Vector3(0, 0, 0)
+
+var loaded_chunks := []  # chunks that are *actually* loaded
+var chunks_to_load := []  # chunks that should be loaded
+var are_chunks_loaded: bool = false setget , get_are_chunks_loaded
+
+var active_chunk = null  # chunk the player is currently in (Vector3)
```
HaSa1002:
```suggestion
var active_chunk: Vector3 = null  # chunk the player is currently in (Vector3)
```
I know `Vector3.ZERO` is a problem probably, but please give it a type hint.

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+
+# needed for serialisation
+func chunk_to_string(chunk: Vector3) -> String:
+	return "%s,%s" % [chunk.x, chunk.z]
+
+
+# needed for deserialisation
+func string_to_chunk(s: String) -> Vector3:
+	var floats = s.split(",", false)
+	return Vector3(int(floats[0]), 0, int(floats[1]))
```
HaSa1002: This is actually dangerous. Could causes OOB. Wouldn't be an issue with TSCN Chunks
```suggestion
# needed for serialisation
func chunk_to_string(chunk: Vector3) -> String:
	return "%s,%s" % [chunk.x, chunk.z]


# needed for deserialisation
func string_to_chunk(s: String) -> Vector3:
	var floats = s.split(",", false)
	if floats.size() <= 1:
		Logger.err("Failed to deserialise chunk position", s)
		return ???
	return Vector3(int(floats[0]), 0, int(floats[1]))
```

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+func is_position_in_loaded_chunk(position: Vector3) -> bool:
+	return is_chunk_loaded(position_to_chunk(position))
+
+
+func is_chunk_loaded(chunk: Vector3) -> bool:
+	return chunk in loaded_chunks
+
+
+func set_position_provider(spatial: Spatial):
+	position_provider = spatial
```
HaSa1002: Why needed? Did you plan to notify something?

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+	]
+
+
+func _ready():
+	grass_mesh = PlaneMesh.new()
+	grass_mesh.size = Vector2(500, 500)
+	grass_mesh.material = preload("res://Resources/Materials/Grass_new.tres")
+
+	if world == null:
+		world = get_parent()
```
HaSa1002: When is world not null?

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+
+
+func _process(delta: float):
+	if world == null:
+		world = get_parent()
+		if world == null:
+			Logger.err("World not found. Freeing self.", self)
+			_kill_thread = true
+			_thread_semaphore.post()
+			queue_free()
```
HaSa1002: That feels like a bug. Why would the world be null?

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+
+	# get position of active camera
+	position_provider = get_viewport().get_camera()
+	if position_provider == null:
+		return
```
HaSa1002: Do we really need the automagic here?

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+
+func _shift_world_origin_to(position: Vector3):
+	_chunk_mutex.lock()
+
+	var delta: Vector3 = position - world_origin
+	world_origin = position
+
+	# TODO: nicer way to check this?
+	if position_provider.find_parent("Players") == null:
+		position_provider.translation += delta
```
HaSa1002: What do you want to check?

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+	for group in groups:
+		var parent = world.get_node_or_null(group)
+		if parent == null:
+			continue
+		for child in parent.get_children():
+			if child == position_provider:
+				continue
+			child.translation += delta
+			if child.has_method("update"):
+				child.update()
```
HaSa1002: What about?
```suggestion
	emit_signal("shif_world_origin", delta)
```

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+
+### SAVING AND LOADING CODE ###
+func _save_chunks(chunks: Array):
+	_chunk_mutex.lock()
+	for chunk in chunks:
+		if chunk in loaded_chunks:
+			_save_chunk(chunk)
+	_chunk_mutex.unlock()
+
+func _save_chunk(chunk_pos: Vector3):
```
HaSa1002:
```suggestion


func _save_chunk(chunk_pos: Vector3):
```
I don't like that save funtion at all

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+				"data": track_object.get_data()
+			}
+
+	world.get_node("jSaveModule").save_value(chunk_to_string(chunk_pos), null)
+	world.get_node("jSaveModule").save_value(chunk_to_string(chunk_pos), chunk)
+	Logger.log("Saved Chunk " + chunk_to_string(chunk_pos))
+
+
+func _get_all_chunks() -> Array:
+	var _all_chunks = []
```
HaSa1002:
```suggestion
	var _all_chunks := []
```

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+		for chunk in get_3x3_chunks(pos):
+			_all_chunks.append(chunk)
+	_all_chunks = jEssentials.remove_duplicates(_all_chunks)
+	return _all_chunks
+
+
+# needed for editor only
+func save_and_unload_all_chunks():
+	if not Root.Editor:
+		return
```
HaSa1002:
```suggestion
	assert(Root.Editor)
```
However calls that in a game is a bug. We should catch them

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+	_save_chunks(_get_all_chunks())
+	_chunk_mutex.lock()
+	chunks_to_load = []
+	_unload_old_chunks()
+
+# this is the missing unlock from `save_and_unload_all_chunks()` - needed for Editor purposes
```
HaSa1002:
```suggestion

# this is the missing unlock from `save_and_unload_all_chunks()` - needed for Editor purposes
```

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+func resume_chunking():
+	_chunk_mutex.unlock()
+
+
+### INTERNAL THREAD LOADING CODE ###
+signal _thread_finished_loading
+var _thread
+var _thread_semaphore
+var _chunk_mutex
+var _kill_thread = false
```
HaSa1002: move up and remove the yelling tag.

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+	emit_signal("chunks_finished_loading")
+	_unload_old_chunks()
+
+
+# called deferred from Thread
+func _add_node_to_scene_tree(parent: String, instance: Spatial):
+	# if an object with this name already exists, stop
+	if world.get_node(parent).get_node_or_null(instance.name) != null:
+		instance.queue_free()
+		return
```
HaSa1002: Feels like bug. assert?

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+
+	world.get_node(parent).add_child(instance)
+	instance.owner = world
```
HaSa1002: Why do we set the owners?

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+
+# handle loading chunks from disk on a separate thread to avoid lags
+func _chunk_loader_thread(_void):
+	while true:
+		_thread_semaphore.wait()
+		# yield(get_tree(), "idle_frame")  # uncomment this for debugging
+		if _kill_thread:
+			return
+
+		_chunk_mutex.lock()
```
HaSa1002:
```suggestion
		_chunk_mutex.lock()
		if _kill_thread:
			_chunk_mutex.unlock()
			return

		
```

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+
+		for chunk_pos in chunks_to_load:
+			if chunk_pos in loaded_chunks:
+				continue
+			var chunk: Dictionary = world.get_node("jSaveModule").get_value(chunk_to_string(chunk_pos), {"empty": true})
```
HaSa1002: You are not TS. get_node on a node in SceneTree

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+
+			if chunk == null:
+				Logger.err("Could not load chunk %s" % str(chunk_pos), self)
+				continue
```
HaSa1002: Dead code?

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+				instance.name = chunk["Buildings"][building].name
+				instance.set_mesh(load(chunk["Buildings"][building].mesh_path))
+				instance.transform = chunk["Buildings"][building].transform
+				var surface_arr: Array = chunk["Buildings"][building]["surfaceArr"]
+				if surface_arr == null:
+					surface_arr = []
+				for i in range (surface_arr.size()):
+					instance.set_surface_material(i, surface_arr[i])
+				instance.set_meta("chunk_pos", chunk_pos)
+				call_deferred("_add_node_to_scene_tree", "Buildings", instance)
```
HaSa1002: Please build the complete buildings tree and instanciate at once. Should be faster

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+
+			## TrackObjects:
+			var track_object_prefab: PackedScene = preload("res://Data/Modules/TrackObjects.tscn")
+			for track_object in chunk["TrackObjects"]:
+				var instance: Spatial = track_object_prefab.instance()
+				instance.name = chunk["TrackObjects"][track_object].name
+				instance.set_data(chunk["TrackObjects"][track_object].data)
+				instance.transform = chunk["TrackObjects"][track_object].transform
+				instance.set_meta("chunk_pos", chunk_pos)
+				call_deferred("_add_node_to_scene_tree", "TrackObjects", instance)
```
HaSa1002: Please build the trackObjects tree at once and instanciate at once. Should be faster, too

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+			if chunk.has("empty"):
+				continue
+
+			## Buildings:
+			for building in chunk["Buildings"]:
+				var instance := MeshInstance.new()
+				instance.name = chunk["Buildings"][building].name
+				instance.set_mesh(load(chunk["Buildings"][building].mesh_path))
+				instance.transform = chunk["Buildings"][building].transform
+				var surface_arr: Array = chunk["Buildings"][building]["surfaceArr"]
```
HaSa1002: This code is assuming the data exists in the chunk. Crying for OOB

src/Data/Scripts/Singletons/chunk_manager.gd
```diff
@@ -0,0 +1,366 @@
...
+
+			## Flora is deprecated anyways... I'll skip it
+
+			## TrackObjects:
+			var track_object_prefab: PackedScene = preload("res://Data/Modules/TrackObjects.tscn")
+			for track_object in chunk["TrackObjects"]:
+				var instance: Spatial = track_object_prefab.instance()
+				instance.name = chunk["TrackObjects"][track_object].name
+				instance.set_data(chunk["TrackObjects"][track_object].data)
+				instance.transform = chunk["TrackObjects"][track_object].transform
```
HaSa1002: This code is screaming for OOB as well

src/Data/Scripts/TrackObjects.gd
```diff
@@ -98,29 +98,29 @@ func attach_to_rail(_rail_node: Spatial) -> void:
...
 func _exit_tree() -> void:
 	unattach_from_rail()
 
-func update(_rail_node: Spatial, res_cache := {}) -> void:
-#	print("Loading Rail Attachment..")
+
+func update() -> void:
+	var _rail_node = world.get_node("Rails").get_node_or_null(attached_rail)
+	if _rail_node == null:
+		return
```
HaSa1002, line 124: Bug?

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -525,48 +507,20 @@ func _on_WorldLoading_AllChunks_pressed() -> void:
...
-		$"World Configuration/WorldLoading/RailConfiguration/FromRail".text,
-		$"World Configuration/WorldLoading/RailConfiguration/ToRail".text,
-		$"World Configuration/WorldLoading/IncludeNeighbours".pressed)
-	if chunks == []:
-		return
-	world.save_chunks(chunks)
+	world.chunk_manager._save_chunks(world.chunk_manager._get_all_chunks())
 
 
+# NOTE: this will also save chunks... great
```
HaSa1002, line 523: Just not save then and remove the callback?
Use _text_changed instead?

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -525,48 +507,20 @@ func _on_WorldLoading_AllChunks_pressed() -> void:
...
 func _on_Notes_Save_pressed() -> void:
 	world.save_value("notes", $"World Configuration/Notes/RichTextLabel".text)
```
HaSa1002, line 525: WAIT. Is that a RichTextLabel?

src/addons/jean28518.jTools/jSaveManager/jSaveModule.gd
```diff
@@ -75,7 +75,7 @@ func _load_current_config():
...
 	_config.load(save_path)
 
-
-func _exit_tree():
-	if _value_changed:
-		write_to_disk()
+# do not do this! user can exit without saving!!
+#func _exit_tree():
+#	if _value_changed:
+#		write_to_disk()
```
HaSa1002: Just remove pls.

src/Data/Scripts/Rail.gd
```diff
@@ -197,7 +197,7 @@ func update_with_calculated_data(calculated_data: Dictionary) -> void:
 		$Ending.transform = get_local_transform_at_rail_distance(length)
 		$Mid.transform = get_local_transform_at_rail_distance(length/2.0)
 		for track_object in trackObjects:
-			track_object.update(self)
+			track_object.update()
```
HaSa1002, line 200: The Rail shouldn't know the TrackObjects. Just do an `emit_signal("rail_changed", whatever)`

HaSa1002 (changes requested):
Not your mistakes, but let's take this opportunity to fix the issues

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: No, 5000 is a reasonable number. Try it. We might be able to increase it to about 8000, anything above that you WILL get NOTICABLE floating point jitter.
This is because floating points give you 6 to 7 digits to work with, we have 4 digits before the comma, so that leaves us 2 after the comma. Since a unit is 1 meter, this means we only have precision in the Millimeter area.
As soon as you jump to Centimeter precision, any user will be able to see artifacts.

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: no can do, Vector3 is not nullable.

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: This is not a problem, since the save file always saves the chunks with a key of `x,z`.

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: You are right, not needed anymore. Was there because it was supposed to be a singleton at first.

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: I don't know. It never should be. But you know as well as I that with this code base, anything is possible.
How often I received nulls when I didn't expect any is beyond me.

src/Data/Scripts/Singletons/chunk_manager.gd
HaSa1002: Yep I know. I am speaking of manual modifications of the file

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: Automagic?
You mean getting the active camera?
Yes, because else the chunk system breaks when you go in the editor or switch to the free camera.
Unless you want to manually set that every time.

It only makes sense to load chunks around the camera, because that is what the player *sees*

HaSa1002: Yes manually setting would be an option, but I agree that it can be better than manually setting it

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: if player is using the FreeCamera for example, it needs to be translated by world origin shift.
But if the player is driving the train, the train itself gets translated to the new origin, and the camera is attached to the train, so we should NOT shift the camera itself.

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: No. I do not want every node to shift itself. This IS the job of a chunk system. No other piece of code should even know that a chunk system exists.

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: So? We need to save chunks... how else are you gonna do it?
I am deliberately keeping the old .save file here, because TSCN chunks may never be approved.

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: Things need an owner. 
And the owner needs to be world, because else find_node() breaks everywhere (find node only finds owned children, iirc)
and if it wasn't find_node, it was some other function. Something breaks somewhere if world it not the owner.

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: Why would you do that?

src/Data/Scripts/Singletons/chunk_manager.gd
HaSa1002: I see

src/Data/Scripts/Singletons/chunk_manager.gd
HaSa1002: I disagree, because we are creating dependencies in a strange way. Let me visualise that at some point. Signals would be an easy way to break those dependencies. Especially if they get injected

src/Data/Scripts/Singletons/chunk_manager.gd
HaSa1002: well then let's do an assert there

src/Data/Scripts/Singletons/chunk_manager.gd
HaSa1002: ah true.

src/Data/Scripts/Singletons/chunk_manager.gd
HaSa1002: Let, the nodes tell the save chunk function what data to save. The save_chunk function just sticks those dicts together.

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: Would still be an issue with TSCN chunks though

src/Data/Scripts/Rail.gd
DasCapschen: We can do this in a separate refactor.

src/Data/Scripts/Singletons/chunk_manager.gd
HaSa1002: Ah didn't know about the find_node thing. I never used that function. I am *getting* around that 😄

src/Data/Scripts/Singletons/chunk_manager.gd
HaSa1002: Because _kill_thread is set not in the thread.

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: Well, it is ensured to exist in at least an empty form by the save function.

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: Buildings are not a tree. They are individual nodes that are all added in World/Buildings.

src/Editor/Docks/Configuration/Configuration.gd
DasCapschen: Yes, so?

src/Editor/Docks/Configuration/Configuration.gd
DasCapschen: Well, if we don't save it, you lose your notes.
The problem is bad design. Notes are saved in the same file as the chunks, and we can only ever write the full file.

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: Actually, this method is never even used. We can just remove it really

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: oki doki

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: asserted instead

src/Data/Scripts/Singletons/chunk_manager.gd
HaSa1002: Would break anything if pack them into a single chunk_x_z node? Because we would spead things up if we only need to add one node to the active tree and not hundred.

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: Hmm, I understand what you mean. 
Perhaps this would be best solved by creating a base class for all Signals / Rails / TrackObjects / etc. which handles the world_origin_shifted signal?
Because for example the parent nodes of Rails etc, which we only use in the tree for order, must not be shifted, only their children.
But I do not want to write code for this in every single class that needs to be shifted.

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: Yeah, unfortunately, we use it a ton in the project 😅

src/Data/Scripts/Singletons/chunk_manager.gd
HaSa1002: Cases (in the chunk we are loading):
1. Missing buildings
2. `Buildings : [ [] ]` (empty building, but entry
3. missing name, mesh_path, transform, and/or surfaceArr

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: hmm, okay.

src/Editor/Docks/Configuration/Configuration.gd
HaSa1002: Ah true, Yeah nevermind

src/Editor/Docks/Configuration/Configuration.gd
HaSa1002: Why though? I think it should be a TextEdit

src/Data/Scripts/Rail.gd
HaSa1002: I would like that

src/Data/Scripts/Singletons/chunk_manager.gd
HaSa1002: Yeah that sounds reasonable. A WorldObject?

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: I can try this, but I think it has a high chance of breaking.

HaSa1002: What about doing it in a subsequent PR/commit?

DasCapschen: Will do as separate PR.

src/Editor/Docks/Configuration/Configuration.gd
DasCapschen: RichTextLabel allows formatting? 😅

src/Data/Scripts/Singletons/chunk_manager.gd
HaSa1002: Works fine with 16 000

src/Data/Scripts/Singletons/chunk_manager.gd
DasCapschen: I think I'll leave the save fuction as is. The Rails don't save any data, just their name.
The buildings are mesh instances, I would have to attach a script to them to get a save function from them.
And at that point, why bother with the TrackObject? It already has most if it's data in get_data(), except name and transform.

HaSa1002 (approved):
lgtm

src/Data/Scripts/chunk_manager.gd
```diff
@@ -0,0 +1,353 @@
+class_name ChunkManager
+extends Node
+
+signal chunks_finished_loading   # emitted when chunks finished loading (loaded_chunks == chunks_to_load)
+signal _thread_finished_loading
+
+const world_origin_shift_treshold: int = 8_000  # if further than this from origin, recenter world origin
```
HaSa1002:
```suggestion
const world_origin_shift_treshold: int = 7_000  # if further than this from origin, recenter world origin
```
because we could be at 7999 on chunk change and thus get into precision issues even then

HaSa1002 (changes requested):
sorry once again...

src/Data/Scripts/chunk_manager.gd
HaSa1002: I just solved the issue by myself :)

HaSa1002 (approved):
lgtm the 2nd

HaSa1002:
Thanks a lot

PR #263, opened 2021-11-21
Fix Shadow Acne

### LINKED ISSUE #254: Directional Light Shadow Acne
Current directional light produces ugly results on our grass plane. We should probably increase some bias values to fix this.

HaSa1002:
Adds LightTools for more realistic light colors. Happens to be an Unlicense licensed library. ¯\\_(ツ)_/¯
Fixes #254


DasCapschen (approved):
lgtm

PR #264, opened 2021-11-21
Fix screenshots not being loaded in builds

HaSa1002:
Caused, because screenshot.png does not exist in packs even though the resource exists.


DasCapschen (approved):
lgtm

PR #266, opened 2021-11-21
Fix a few bugs

### LINKED ISSUE #171: Deselection with click into void
When you have an object selected in the editor and you click somewhere else, you should deseselect that object.

-> use unhandled_input()

### LINKED ISSUE #242: Editor does not find all resources in exported Project

HaSa1002:
- Fix content selector not displaying resources in builds (Fixes #242)
- Fix RailAttachments causing script error because of TrackObject.update
- Fix signals (the ingame one) causing script error because of types
- Remove select collider ~~(without replacement for now. The objects should provide their own collider)~~
- Replaced select collider with dedicated colliders
- Reworked selection method to allow deselection with clicking on nothing (Convienently fixes #171)
- Adds AABB to collider helper script


src/Data/Scripts/Singletons/Root.gd
```diff
@@ -111,25 +111,22 @@ func crawlDirectory(directoryPath: String, foundFiles: Array, fileExtensions: Ar
...
-				else:
-					exportString = directoryPath +"/"+file
-				foundFiles.append(exportString)
+			crawlDirectory(directoryPath.plus_file(file), foundFiles, fileExtensions)
+			continue
+
+		var ext := file.get_extension()
+		if OS.has_feature("standalone") and ext == "import":
+			file = file.get_basename()
+			ext = file.get_extension()
```
DasCapschen, line 127: Was the problem really the .import files? 🤔 
It used to work when we didn't load multiple mods...

But I guess this filters out the .import files from the file list?

src/Editor/Scripts/Editor.gd
```diff
@@ -312,13 +304,6 @@ func load_world() -> bool:
 	if last_editor_camera_transforms.has(current_track_name):
 		camera.load_from_transform(last_editor_camera_transforms[current_track_name])
 
-	## Add Colliding Boxes to Buildings:
-	for building in $World/Buildings.get_children():
-		building.add_child(preload("res://Editor/Modules/SelectCollider.tscn").instance())
-	for signal_ins in $World/Signals.get_children():
-#		if signal_ins.type == "Signal":
-		signal_ins.add_child(preload("res://Editor/Modules/SelectCollider.tscn").instance())
```
DasCapschen, line 320: - We won't be able to select any Objects or Signals. That's not good. Why did you remove the select collider at all? What was the problem with it?

- >The objects should provide their own collider

  - Can you at least provide colliders for Signals then?

src/Resources/SignalTypes/Hp/Form/HpForm.tscn
```diff
@@ -117,10 +117,6 @@ script = ExtResource( 2 )
 [node name="formsignal_hp" parent="." instance=ExtResource( 1 )]
 transform = Transform( 1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 1.667 )
 
-[node name="Skeleton" parent="formsignal_hp/Armature" index="0"]
-bones/0/bound_children = [ NodePath("lowerLightAttachment") ]
-bones/1/bound_children = [ NodePath("upperLightAttachment") ]
-
```
DasCapschen: This will break Signal animations.
I'm guessing Godot removed this by itself (though I don't understand why, it does not for me). 
Please fix though.

DasCapschen (changes requested):
Please consider comments. Other than that, looks good.

src/Data/Scripts/Singletons/Root.gd
HaSa1002: > Was the problem really the .import files? 🤔
> It used to work when we didn't load multiple mods...

yes it is. It was always broken.

> But I guess this filters out the .import files from the file list?

C'mon. That is not a valid use case.

src/Editor/Scripts/Editor.gd
HaSa1002: They don't work with signals as well. I just have to put colliders onto the signals itself. I tend to push towards the same direction for the buildings and just use them as scenes

src/Data/Scripts/Singletons/Root.gd
DasCapschen: >yes it is. It was always broken.

Ah, alright then. Still kinda weird, how it worked before then.

>C'mon. That is not a valid use case.

I meant your code, what you're doing now is basically changing the path so it points to the original, instead of the .import, so we shouldn't see any .import files in the list.

HaSa1002: Yes that is the case. Because the loader loads the "non existent" file using the resource path

src/Editor/Scripts/Editor.gd
DasCapschen: Yeah, I get that you removed the select collider because they are broken. But can you add colliders to the Signals scenes itself then, like you suggested? So we can select them again?

HaSa1002: yeah will do tomorrow or the day after or when I have time again :D
Soon™

src/Data/Scripts/Singletons/Root.gd
DasCapschen: Alright. Sounds good then.

src/Editor/Scripts/Editor.gd
HaSa1002: fixed

HaSa1002:
Added more fixes 🙈 Sorry for bloating this one. This PR is insanly packed now

DasCapschen (approved):
lgtm

PR #271, opened 2021-11-30
fix crawl directory

DasCapschen:


src/Data/Scripts/Singletons/Root.gd
```diff
@@ -105,62 +105,63 @@ func checkAndLoadTranslationsForTrain(trainDirPath: String) -> void:
 		TranslationServer.add_translation(tainTranslation)
 
 
-func crawlDirectory(directoryPath: String, foundFiles: Array, fileExtensions: Array) -> void:
+# recursion_limit = -1 -> "unlimited" recursion (-int.max)
+# recursion_limit = 0 -> no recursion, only search in directory_path
+# recursion_limit = 2 -> this directory, subdirectories, and subdirectories of subdirectories
+# found_files is return value
+func crawl_directory(found_files: Array, directory_path: String, file_extensions: Array, recursion_limit: int = -1) -> void:
```
HaSa1002: why isn't the recursion limit the number of subdirectories?

src/Data/Scripts/Singletons/Root.gd
```diff
@@ -105,62 +105,63 @@ func checkAndLoadTranslationsForTrain(trainDirPath: String) -> void:
...
-
-
 func fix_frame_drop() -> void:
 	if not jSettings.get_framedrop_fix():
 		return
 	jEssentials.call_delayed(0.7, self,  "set_fullscreen", [!jSettings.get_fullscreen()])
 	jEssentials.call_delayed(0.9, self,  "set_fullscreen", [jSettings.get_fullscreen()])
 
 
+func make_filename_valid(filename: String) -> String:
```
HaSa1002: i would like `to_valid_filename` more.

HaSa1002 (commented):
some thoughts

src/Data/Scripts/Singletons/Root.gd
DasCapschen: huh? the recursion limit *is* the number of subdirectories :thinking:
0 subdirectories -> only this directory
1 -> 1 level of subdirs
2 -> 2 levels of subdirs
-1 -> unlimited (because that was the old behaviour)

src/Data/Scripts/Singletons/Root.gd
DasCapschen: done

src/Data/Scripts/Singletons/Root.gd
HaSa1002: That comment is misleading then. Because to me it looked like a toggle behaviour. Probably better to just rename the `recursion_limit` to `subfolder_limit` or `subfolder_depth` and remove the comments

HaSa1002: well the second part of the suggestion is actually better than removing the comments.

> Probably better to just rename the recursion_limit to subfolder_limit or subfolder_depth

DasCapschen: The name is fine as it is. Because it says exactly what it is. It is the recursion limit.

HaSa1002: Yeah that is the case. But in a search algorithmn traversing the file system, a recursion limit is also a depth. Hence my second proposal 🙈. But feel free to leave as it is and squash the commits :) Will approve and merge after squash

DasCapschen: It's now recursion_depth. Squashed.

HaSa1002 (approved):
lgtm

HaSa1002:
Thanks 🎉

PR #275, opened 2021-12-04
Enable MSAA x4 for JFR1 screen textures

HaSa1002:
This should increase the readability of the screens.


DasCapschen:
It *should*, but does it?
Also what kind of performance impact does this have?

HaSa1002:
I feel so. I don't know. There is not notable difference on mt RTX 3060 😅

DasCapschen (approved):
lgtm

PR #276, opened 2021-12-05
fix chunks not unloading correctly

DasCapschen:


src/Data/Scripts/Rail.gd
```diff
@@ -146,59 +117,53 @@ func calculate_update() -> Dictionary:
...
+		multimeshI.multimesh.transform_format = MultiMesh.TRANSFORM_3D
+		multimeshI.multimesh.set_meta("is_unique", true)
+
+	multimeshI.multimesh.instance_count = visibleSegments
+
+	# this is called, but mesh remains null
+	if multimeshI.multimesh.mesh == null:
+		multimeshI.multimesh.mesh = rail_type_node.get_child(0).mesh
+		for i in range(rail_type_node.get_child(0).get_surface_material_count()):
+			multimeshI.multimesh.mesh.surface_set_material(i, rail_type_node.get_child(0).get_surface_material(i))
```
HaSa1002, line 151: I noticed, sometimes checking == null doesn't work if instances were freed. Only `is_instance_valid` works there. But issue may be different.

HaSa1002 (approved):
lgtm

src/Data/Scripts/Rail.gd
HaSa1002: To be clear: Noticed in a different prototype

HaSa1002:
Thank you :)

PR #295, opened 2022-01-02
Improe cabin inside camera script

Jean28518:
- Reverts the direction of the mouse wheel zoom inside the cabin. -> That's more familiar with zoom in and zoom out
- Don't hardscript anymore some camera fov settings
- Enable camera movement with mouse only, if right mouse button is clicked.

### DIFF
--- a/src/Data/Scripts/Player.gd
+++ b/src/Data/Scripts/Player.gd
@@ -84,11 +84,11 @@ var has_camera_distance_changed: bool = false
 const CAMERA_DISTANCE_MIN: float = 5.0
 const CAMERA_DISTANCE_MAX: float = 200.0
 
-var ref_fov: float = 42.7 # reference FOV for camera movement multiplier
-var camera_fov: float = 42.7 # current FOV
-var camera_fov_soll: float = 42.7 # FOV user wants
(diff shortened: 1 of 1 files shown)


src/Data/Scripts/Player.gd
```diff
@@ -691,14 +691,15 @@ func handleCamera(delta: float) -> void:
 		cameraNode.translation = soll_camera_translation
 
 		# FIXME: in the first frame, delta == 0, why?
-		if mouseMotion.length() > 0 and delta > 0:
+		if mouseMotion.length() > 0 and delta > 0 and (Input.is_mouse_button_pressed(BUTTON_RIGHT) or Root.mobile_version):
```
DasCapschen, line 694: Why the Root.mobile_version ? :thinking:

DasCapschen (approved):
Looks great, I have no objections 👍🏻

HaSa1002 (approved):
lgtm

src/Data/Scripts/Player.gd
HaSa1002: Doesn't make sense. Let's do merge and fix later as it shouldn't break anything.

PR #304, opened 2022-01-09
Fix state after scenario editor

### LINKED ISSUE #301: Track Editor: World doesn't load when you edited something in the scenario editor before.
world.chunk_loader is null.

Jean28518:
I did forget to reset the states in the Root script when exiting scenario editor

Fixes #301


DasCapschen (approved):
lgtm

PR #305, opened 2022-01-09
Add: Selection of buildings in the editor

### LINKED ISSUE #288: Buildings are not selectable anymore in the track editor

Jean28518:
Fixes #288

### DIFF
--- a/src/Data/Scripts/chunk_manager.gd
+++ b/src/Data/Scripts/chunk_manager.gd
@@ -241,6 +241,7 @@ func _add_node_to_scene_tree(parent: String, instance: Spatial):
 	if instance.has_method("update"):
 		instance.update()
 
+
 func append_deduplicated(A: Array, B: Array):
 	for b in B:
 		if not b in A:
@@ -276,7 +277,7 @@ func _unload_old_chunks(all: bool = false):
 	for group in ["TrackObjects", "Buildings", "Landscape"]:
 		var parent = world.get_node(group)
 		for child in parent.get_children():
-			if child.get_meta("chunk_pos") in chunks_to_unload:
+			if all or (child.get_meta("chunk_pos") in chunks_to_unload):
 				child.free()
 
 	# remove unloaded chunks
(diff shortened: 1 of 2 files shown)


src/Data/Scripts/chunk_manager.gd
```diff
@@ -323,6 +324,10 @@ func _chunk_loader_thread(_void):
 					for i in range (surface_arr.size()):
 						instance.set_surface_material(i, surface_arr[i])
 					instance.set_meta("chunk_pos", chunk_pos)
+					instance.set_script(preload("res://Data/Scripts/aabb_to_collider.gd"))
+					instance.target = NodePath(".")
+					instance.generate_collider()
+					instance.set_script(null)
```
DasCapschen: I think this would be better, in case the object already has a script.
Although, I don't really like that we're modifying script at run time at all 🤔 
```suggestion
					var old_script = instance.get_script()
					instance.set_script(preload("res://Data/Scripts/aabb_to_collider.gd"))
					instance.target = NodePath(".")
					instance.generate_collider()
					instance.set_script(old_script)
```

src/Editor/Scripts/Editor.gd
```diff
@@ -446,6 +446,10 @@ func add_object(complete_path: String) -> void:
 	mesh_instance.translation = position
 	$World/Buildings.add_child(mesh_instance)
 	mesh_instance.set_owner($World)
+	mesh_instance.set_script(preload("res://Data/Scripts/aabb_to_collider.gd"))
+	mesh_instance.target = NodePath(".")
+	mesh_instance.generate_collider()
+	mesh_instance.set_script(null)
```
DasCapschen: same as above, probably better to save a potential old script before overwriting it.

DasCapschen (changes requested):

src/Editor/Scripts/Editor.gd
Jean28518: Done!

Jean28518:
Can we merge this one? :)

PR #308, opened 2022-01-09
Fix visibilty update of npc trains in unloaded chunks

### LINKED ISSUE #286: Fix visibility of npc trains
When they spawn in an unloaded chunk, they never appear are crossing you

Jean28518:
Fixes #286


DasCapschen (approved):
lgtm

PR #316, opened 2022-01-22
Fix chunk manager at loading far away from world origin

Jean28518:
It makes sense to store this value in .tscn because all rail positions are moved too in the .tscn file.


HaSa1002 (approved):
lgtm

PR #317, opened 2022-01-22
Fix editor camera not grabbing keyboard focus

### LINKED ISSUE #300: Track Editor: Keys are inserted while flying. After editing any text field
grafik

HaSa1002:
Fixes #300

Focus is given back to control that was focused previously when the fly mode is exited.

### DIFF
--- a/src/Editor/Scripts/editor_camera.gd
+++ b/src/Editor/Scripts/editor_camera.gd
@@ -6,12 +6,15 @@ export var mouse_sensitivity: float = 0.003
 export var normal_speed: float = 1
 export var fast_speed: float = 3
 export var pan_move_sensitivity: float = 0.25
+export var control_path: NodePath
 
 
 var velocity := Vector3.ZERO
 var is_moving_first_person := false
 var is_panning := false
 var in_movement_time: float = 0
+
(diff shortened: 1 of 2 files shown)


src/Editor/Scripts/editor_camera.gd
```diff
@@ -77,3 +80,15 @@ func _physics_process(delta: float) -> void:
...
+	._capture_mouse()
+	var focus_owner := control.get_focus_owner()
+	if focus_owner:
+		focus_owner.release_focus()
+		control = focus_owner
+
+
+func _free_mouse() -> void:
+	._free_mouse()
+	control.grab_focus()
```
DasCapschen, line 94: could control be null?

DasCapschen (approved):
lgtm

Jean28518 (approved):

src/Editor/Scripts/editor_camera.gd
HaSa1002: Only if it is not assigned

HaSa1002: in the editor

PR #319, opened 2022-01-22
Remove content import feature

### LINKED ISSUE #313: Bug: Import of .pck files doesn't work
grafik

How to reproduce:

content -> Import -> Select a .pck file

HaSa1002:
Fixes #313

Feature is not supportable for 0.9

### DIFF
--- a/src/Data/UI/Content.gd
+++ b/src/Data/UI/Content.gd
@@ -23,28 +23,5 @@ func _on_Back_pressed() -> void:
 	hide()
 
 
-func _on_Import_pressed():
-	$FileDialog.current_path = OS.get_system_dir(OS.SYSTEM_DIR_DOWNLOADS)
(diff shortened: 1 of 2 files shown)


Jean28518:
? How to install addons easily?

DasCapschen:
You need both a .pck file with the addon, and a mod config (content.tres).
Simply copying the .pck file is not enough. If you want to make a "import content" feature, either you need to copy a whole folder, containing both, or you could distribute addons as .zip files, which this method must then unpack into a subfolder in addons/

```
addons/
|---- my_addon/
       |---- content.tres
       |---- mod.pck
```


Also the function `ContentLoader.update_config()` does not even exist.

Jean28518:
Thanks. For v1.0 we should implement such a feature :)

Jean28518 (approved):

HaSa1002:
There is no Zip feature in Godot. At least none that has a GDScript interface

PR #320, opened 2022-01-22
Fix translations not getting applied

### LINKED ISSUE #297: Translation 'MENU_FEEDBACK_QUESTION' not available
grafik

Many other translations are also not available anymore...

HaSa1002:
Fixes #297

RichTextLabels don't translate. I also added some translation files back and enabled some translation that we have.


DasCapschen (approved):
lgtm

PR #323, opened 2022-01-23
Updated the Bulgarian and English translations

RacerBG:
Disclaimer: I'm not a native English speaker so do not expect miracles from my work but I believe that the word order of the English sentences is better now.

And also I had translated the latest changes in to Bulgarian as well.


Jean28518 (approved):
Thanks for your contribution, looks fine :)

PR #341, opened 2022-02-13
Add Debug UI (FPS Counter)

### LINKED ISSUE #340: Show FPS in-game
Would it be possible to have an option in Settings to turn on an FPS counter?

My laptop is struggling with the frame rate, and it would be nice to have an FPS counter while adjusting the graphics options.

DasCapschen:
Adds an FPS Counter in the upper right.
Fixes #340 

Can be activated by pressing `F12` at any time. (uses a keybinding, so can be rebound later)

grafik


HaSa1002 (changes requested):
FYI, file names in Godot should be lower_case style.
Please prefer  `_unhandled_input` over `_input` whenever possible.

DasCapschen:
My files are in `lower_case` style ... ❓ 

I think this situation calls for _input.
Its a debug interface, you don't want anything else blocking it from opening.

HaSa1002:
> My files are in lower_case style ... ❓

My bad. I was referring to the `misc` folder but apperantly there is already one.

> I think this situation calls for _input.
Its a debug interface, you don't want anything else blocking it from opening.

I disagree because of the reasons I posted in our private conversation.

DasCapschen:
Changed to `_unhandled_input` @HaSa1002

HaSa1002:
Thanks. Please squash the last two commits together 😇

DasCapschen:
Done. Though you could've just squash-merged 😛

HaSa1002 (approved):
lgtm. No, because first, I don't like squash merges and second, the movement of the logo files should be a seperate commit

HaSa1002:
thanks ❤️

PR #343, opened 2022-02-20
Add content (vegetation, train stations, bridges, undercrossings, transmission tower)

Jean28518:
Adds some content for more variation :)


HaSa1002 (approved):
ok, I guess. I do have a problem with the file names but will we fix it for 1.0

PR #347, opened 2022-03-19
Fix Advanced Train Driving scenario name

rogerm4242:
Renamed 'Advanced Train Driving.scenerio'.


DasCapschen (approved):
Looks good, thank you for your contribution :+1:

PR #353, opened 2022-04-14
Make text_box use the new theme

nalquas:
Currently, text boxes (as used in the tutorials and to show scenario descriptions on start) use the Godot theme instead of the custom theme used in other parts of the GUI. This pull request fixes that.

I also needed to change the text_box script to use `grab_click_focus()` instead of `grab_focus()` to get the OK button to be readable (not white on white) without first having to click on the message panel.

Before:
Screenshot_20220414_203744

After (without focus grab fix):
Screenshot_20220414_204154

After (with focus grab fix):
Screenshot_20220414_204305

### DIFF
--- a/src/Data/UI/text_box.gd
+++ b/src/Data/UI/text_box.gd
@@ -18,7 +18,7 @@ func message(text: String) -> void:
 	$MarginContainer/VBoxContainer/Message.text = text
 	visible = true
 	get_tree().paused = true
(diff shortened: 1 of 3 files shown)


HaSa1002 (approved):
lgtm

DasCapschen:
Looks great, thank you! ❤️

PR #356, opened 2022-04-16
Workaround Godot Editor resetting chunk_unload_distance to 0

nalquas:
Sorry, one more workaround before going to sleep.
Fixes this problem, caused by Godot's Editor resetting chunk_unload_distance to 0 on launch:
Screenshot_20220417_015230

### DIFF
--- a/src/addons/jean28518.jTools/jSettings/JSettings.gd
+++ b/src/addons/jean28518.jTools/jSettings/JSettings.gd
@@ -15,7 +15,11 @@ func _ready():
 
 func apply_saved_settings():
 	if ProjectSettings["game/debug/first_run"]:
+		# Workaround Godot resetting default settings in Editor
 		set_fullscreen(true)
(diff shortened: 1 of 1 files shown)


HaSa1002 (approved):
lgtm

HaSa1002:
Thanks ❤️

nalquas:
Ah, I'm now sure the whole project settings changing-situation is a Godot bug. Apparently, it will be fixed in 3.5, as per this issue https://github.com/godotengine/godot/issues/30912 and this pull request https://github.com/godotengine/godot/pull/55230.

So as soon as Godot 3.5 is out, we should be able to remove this workaround. The workaround, by the way, seems to be affected by the issue as well, so it's a bit flaky.
If you release v0.9 before Godot 3.5 comes out, please make *sure* to set the `debug/first_run` variable to `true` before export, or the workaround will never activate.

PR #358, opened 2022-05-05
Apply default settings if override.cfg doesn't exist yet

nalquas:
Instead of relying on a first_launch variable to ensure correct initial settings, simply checking for game-breaking settings on launch should be enough to avoid game-breaking config issues. 

Some points:

- This seems to be a more robust solution than [!356](https://github.com/Libre-TrainSim/Libre-TrainSim/pull/356), and should prevent future issues in this regard. Obviously, chunk_unload_distance shouldn't ever have a value of 0 in the first place, but I'd keep the sanity check in, just in case.
- This change has the side effect of the game launching in windowed mode by default, though I don't see that as a problem.
- The first_launch variable shouldn't be needed in the first place, and assuming correct export procedure, the default values (such as SIFA or PZB) appear to be loaded correctly.
- The issue of Godot Editor overwriting `project.godot` on every launch based on what's written in the developer's user directory (`~/.local/share/libre_train_sim/override.cfg`) remains, though that should be fixed by Godot upstream (see https://github.com/godotengine/godot/issues/30912). Until then, assuming whoever does the final export of the game deletes their `override.cfg` and reverts any changes to their `project.godot` before export (plus maybe restarts Godot), everything should be fine.
- Even if the old workaround would have worked, the `first_launch` variable itself was affected by the bug, thus making it unreliable.

### DIFF
--- a/src/addons/jean28518.jTools/jSettings/JSettings.gd
+++ b/src/addons/jean28518.jTools/jSettings/JSettings.gd
@@ -17,7 +17,7 @@ func _ready():
 func first_run_check():
 	# Check if this is the first run, and if it is, apply default settings
 	var dir = Directory.new()
-	if not dir.file_exists("user://override.cfg"):
+	if not dir.file_exists("user://override.cfg") or OS.has_feature("editor"):
 		Logger.log("First run (\"user://override.cfg\" doesn't exist). Applying default settings.")
 		set_fullscreen(true)
 		set_vsync(true)
--- a/src/addons/jean28518.jTools/jSettings/JSettings.gd
+++ b/src/addons/jean28518.jTools/jSettings/JSettings.gd
@@ -43,7 +43,9 @@ func apply_saved_settings():
 
 
 func save_settings():
-	ProjectSettings.save_custom("user://override.cfg")
+	# Save, except if in Godot Editor
+	if not OS.has_feature("editor"):
+		ProjectSettings.save_custom("user://override.cfg")
 
 
 func update_settings_window():
--- a/src/addons/jean28518.jTools/jSettings/JSettings.gd
+++ b/src/addons/jean28518.jTools/jSettings/JSettings.gd
@@ -10,19 +10,33 @@ func popup():
 func _ready():
 	if get_parent().name == "root":
 		$JSettings.hide()
+	first_run_check()
 	apply_saved_settings()
 
 
-func apply_saved_settings():
-	if ProjectSettings["game/debug/first_run"]:
-		# Workaround Godot resetting default settings in Editor
+func first_run_check():
+	# Check if this is the first run, and if it is, apply default settings
+	var dir = Directory.new()
+	if not dir.file_exists("user://override.cfg"):
+		Logger.log("First run (\"user://override.cfg\" doesn't exist). Applying default settings.")
 		set_fullscreen(true)
+		set_vsync(true)
+		set_anti_aliasing(2)
+		set_fog(true)
+		set_shadows(true)
+		set_main_volume(1.0)
+		set_music_volume(1.0)
+		set_game_volume(1.0)
+		set_persons(true)
+		set_view_distance(1000)
 		set_sifa(true)
 		set_pzb(true)
-		set_chunk_unload_distance(2.0)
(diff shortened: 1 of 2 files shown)


HaSa1002:
> This change has the side effect of the game launching in windowed mode by default, though I don't see that as a problem.

I feel like we should just check if a `user://overrides.cfg` exists and set the variables including fullscreen there. This is probably faster and doesn't require maintenance of checkworthy variables. What do you think?

nalquas:
> I feel like we should just check if a `user://overrides.cfg` exists and set the variables including fullscreen there.

I agree, it's certainly better than having a `first_launch` variable.

>This is probably faster and doesn't require maintenance of checkworthy variables. What do you think?

But then we would instead have to maintain a (redundant) list of default settings values in our code. Since the whole point of using `ProjectSettings` instead of `JSettings`/`JSaveManager` was to have less overhead, this might be counterproductive.

The default settings are, after all, saved in `project.godot`. The problem is that Godot Editor overwrites them at startup with whatever is saved in `user://overrides.cfg`. This can lead to weird issues for developers, mainly that certain settings won't be saved correctly as Godot wants to save space by only saving stuff that's changed from the default settings. Since the default settings are erroneously overwritten (as described), your changes are suddenly no longer different from the (new/wrong) default, thus leading to them not being saved.

This problem shouldn't affect exported builds of the game, though, and seems to be a developer-only problem.

I really only see two options, then:
Either we keep setting default settings at first launch manually (though now by checking for whether `user://overrides.cfg` exists) and accept the redundancy, or we remove the first launch check and the sanity check entirely. I'm not sure how to proceed, what's your opinion?

HaSa1002:
Hmm. Too bad this is only fixed in 3.5. IMHO I'd go with two things for the time being. First, let's add an overwrite for the fullscreen mode and second, disable saving settings in editor builds. This is not ideal but until we have 3.5 (which shouldn't be too long in the future), I don't really see any other solution. Your points regarding redundancy are valid but this is happens to be only an issue for a small number of settings that I would bite the bullet until 3.5 is out. I am pretty sure we are going to have 3.5 in the coming one or two months.

Besides why did that happen anyway?
> This change has the side effect of the game launching in windowed mode by default, though I don't see that as a problem.

HaSa1002:
Actually just put set_fullscreen into the chunk unload distance check as well and I am happy, I guess.

HaSa1002:
No need to overcomplicate the thing. (Alternately you can change to check for the existence of the file and set fullscreen there)

nalquas:
> second, disable saving settings in editor builds.

Done. I solved this by changing the user:// location for the editor only, and preventing the game from writing an override.cfg in that case. This definitely stops Godot from overwriting the `project.godot` (which is nice).

> you can change to check for the existence of the file and set fullscreen there

Done. All our custom settings are now set in the `first_run_check()` function.

HaSa1002:
> Done. I solved this by changing the user:// location for the editor only, 

I am not sure this is a good idea. I mean this is only noticed by us few devs, but this practically means that the ingame editor data has to be copied when switching between editor and standalone builds. @DasCapschen wdyt?
Isn't the save custom check enough?

> and preventing the game from writing an override.cfg in that case. This definitely stops Godot from overwriting the project.godot (which is nice).

Great. Also sorry for the back and forth. The smaller the change the easier is nitpicking. Otherwise this is totally good to go

nalquas:
> > Done. I solved this by changing the user:// location for the editor only,
>
> I am not sure this is a good idea. I mean this is only noticed by us few devs, but this practically means that the ingame editor data has to be copied when switching between editor and standalone builds.

I did a small change (ee7e8414149bd5e44aad232fd9b5a41041ccb9d4) which should have the same effect, but without a seperate `user://` directory. I've pointed Godot Editor at a non-existent `user://override_dev.cfg`. So saving stays disabled for devs, but at least the user directory is shared with exported builds of the game.

HaSa1002 (approved):
lgtm. I'd say we keep the history for future reference

HaSa1002:
Thank you very much ❤️

HaSa1002:
Cherry picked for 0.9.beta2

PR #362, opened 2022-05-14
Overhaul the material selector/editor

HaSa1002:
While testing we found a ton of issues with the picker. This is not the major UI improvement but at least enables editing objects.
 * Remove faulty mouse event signals
 * Guess the desired default material from the surface name
 * Display the surface name if the material cannot be guessed
 * Fix pressing the update button causes a rebuild of the UI and thus
   leading to setting non-existent surfaces
 * Improve code structure- Fix editor camera freed instance issues when modifying focus
- Overhaul the objects selector
- Rename BuildingSettings to building_settings

grafik
grafik


HaSa1002:
Please review the commits separate from each other. There are so many changes that git doesn't noticed the rename is a rename.

src/Editor/Modules/building_settings.gd
```diff
@@ -0,0 +1,140 @@
...
+	for folder in ContentLoader.repo.material_folders:
+		Root.crawl_directory(materials, folder, ["tres", "res"])
+	for material in materials:
+		possible_materials[material.get_file().get_basename()] = material
+
+
+func _input(event):
+	if visible and not is_instance_valid(current_mesh):
+		Logger.warn("current_mesh was deleted but the building settings haven't been updated!", self)
+		hide()
```
DasCapschen, line 30: Why is this on _input ?

src/Editor/Modules/building_settings.gd
```diff
@@ -0,0 +1,140 @@
...
+var possible_materials := {}
+
+func _ready():
+	$"Material-1".hide()
+
+	var materials = []
+	for folder in ContentLoader.repo.material_folders:
+		Root.crawl_directory(materials, folder, ["tres", "res"])
+	for material in materials:
+		possible_materials[material.get_file().get_basename()] = material
```
DasCapschen, line 24: Could this create collisions if there is both a .tres and .res material?

src/Editor/Modules/building_settings.gd
```diff
@@ -0,0 +1,140 @@
...
+			if material_name in possible_materials:
+				current_material = load(possible_materials[material_name])
+				line_edit.text = possible_materials[material_name]
+				current_mesh.set_surface_material(i, current_material)
+			else:
+				line_edit.text = ""
+				line_edit.placeholder_text = material_name
+		else:
+			line_edit.text = ""
+			line_edit.placeholder_text = "Can't guess the material"
```
DasCapschen, line 66: Can this ever happen? Mesh would need to be null...

src/Editor/Modules/building_settings.gd
```diff
@@ -0,0 +1,140 @@
...
+		new_child.show()
+
+	# Add Update Button
+	## TODO: Do we really need that boy?
+	var button := Button.new()
+	button.name = "Button"
+	button.text = "Update"
+	button.connect("pressed", self, "set_current_config_to_mesh")
+	button.connect("pressed", self, "emit_signal", ["updated"])
+	add_child(button)
```
DasCapschen: Can't you just call `set_current_config_to_mesh()` and `emit_signal("updated")` directly?

DasCapschen (commented):
Looks OK, check comments.

src/Editor/Modules/building_settings.gd
HaSa1002: idk. It has been there. I didn't wanna break stuff so I left it there with the warning. I did see it still acting so I guess the editor needs to perform more to ensure this not being needed.

src/Editor/Modules/building_settings.gd
HaSa1002: Yes, but I don't care. If the name is the same, the last material is going to be preselected. I don't see too much harm there.

src/Editor/Modules/building_settings.gd
HaSa1002: Well. In theory the mesh could be a primitive mesh. But if there is an array mesh, you are correct.

src/Editor/Modules/building_settings.gd
HaSa1002: I am not even sure we still need the button. However if you don't pick the material and enter the path directly instead, the material won't be updated. But `emit_signal` is probably a duplicate.

src/Editor/Modules/building_settings.gd
DasCapschen: Primitive meshes have a null mesh attribute?

src/Editor/Modules/building_settings.gd
DasCapschen: Alright, we'll leave it in then.

src/Editor/Modules/building_settings.gd
HaSa1002: No, but they don't have surface names

DasCapschen: So then the "Can't guess the material" else case is dead code? 🤔

HaSa1002: It will get triggered if the assigned mesh is not an array mesh but a primitive mesh

HaSa1002: To explain it more thorough: We get a MeshInstance node there. This node has a mesh. This mesh is of type mesh. Both `ArrayMesh` and `PrimitiveMesh` inherit from from `Mesh`. However, while `ArrayMesh`s do have `surfacee_get_name(int idx)` `PrimitiveMesh`s, as well as their base class have not. That is the only reason why I put it there. Because we still can edit the materials but we can't get the surface name because the API doesn't support it.

DasCapschen: oh right, I didn't see you are casting it to an ArrayMesh. All fine then.

src/Editor/Modules/building_settings.gd
DasCapschen: So we need it for the case someone types it in, instead of using the picker? 
That's fine by me, we can leave the button in then.

HaSa1002: yeah. I improved another little thing there and I am about to push it

HaSa1002:
changes made

DasCapschen (approved):
lgtm

HaSa1002:
Cherry picked for 0.9.beta2

PR #364, opened 2022-05-18
Fix mod tools not creating new mods

HaSa1002:
Also disallow the creation of mods if the unique name is empty as it would create the mod in the parent directory.

Fixes #360

PR #361, opened 2022-05-14
Fix editor camera freed instance issues when modifying focus

HaSa1002:
Since we are deleting controls quite often in the editor, this code was prone to breaking.

PR #359, opened 2022-05-07
Fix duplicate language selector problem

nalquas:
fixes #355 
Screenshot_20220507_132804

PR #365, opened 2022-05-20
Cherry-picks for 0.9 Batch 1 (beta2)

HaSa1002:
Includes #358 #359 #361 #362 #364


HaSa1002:
Oh well... We don't have CI set up for 0.9 and no merge restrictions. I am going to change that after I pressed that button xD

PR #375, opened 2022-06-29
Prefer queue_free() over free() where appropriate

nalquas:
This pull request replaces all remaining uses of `free()` with `queue_free()`.
My reasoning behind this is that `queue_free()`, contrary to `free()`, ensures it's safe to delete a node ([according to Godot Docs](https://docs.godotengine.org/en/stable/classes/class_node.html?highlight=queue_free()#class-node-method-queue-free)), and thus should be more reliable.

Please test for side-effects.

### DIFF
--- a/src/Data/Scripts/Rail.gd
+++ b/src/Data/Scripts/Rail.gd
@@ -174,9 +174,9 @@ func _update() -> void:
 func unload_visible_instance() -> void:
 	visible = false
 	if get_node_or_null("MultiMeshInstance") != null:
-		$MultiMeshInstance.free()
+		$MultiMeshInstance.queue_free()
 	if get_node_or_null("OverheadLine") != null:
-		$OverheadLine.free()
+		$OverheadLine.queue_free()
 	for track_object in trackObjects:
 		if is_instance_valid(track_object):
 			track_object.queue_free()
--- a/src/Editor/Docks/RailBuilder/RailBuilder.gd
+++ b/src/Editor/Docks/RailBuilder/RailBuilder.gd
@@ -175,7 +175,7 @@ func _on_Rename_pressed() -> void:
 
 
 func _on_Delete_pressed() -> void:
-	currentRail.free()
+	currentRail.queue_free()
 	currentRail = null
 	update_selected_rail(self)
 	Logger.log("Rail deleted.")
--- a/src/Editor/Scripts/ScenarioConfiguration.gd
+++ b/src/Editor/Scripts/ScenarioConfiguration.gd
@@ -154,7 +154,7 @@ func update_ui_for_current_route():
 	$TabContainer/Routes/RouteConfiguration/GeneralSettings/P.visible = routes[current_route].general_settings.activate_only_at_specific_routes and not routes[current_route].general_settings.player_can_drive_this_route
 
 	for child in $TabContainer/Routes/RouteConfiguration/GeneralSettings/P/SpecificRoutes.get_children():
-		child.free()
+		child.queue_free()
 	for route_name in routes.keys():
 		if routes[route_name].general_settings.player_can_drive_this_route and route_name != current_route:
 			var checkbox: CheckBox = CheckBox.new()
--- a/src/Editor/Scripts/scenario_map.gd
+++ b/src/Editor/Scripts/scenario_map.gd
@@ -173,9 +173,9 @@ func update_map():
 
 	# Clear old selection:
 	for child in $Special.get_children():
-		child.free()
+		child.queue_free()
 	for child in $RailsSelection.get_children():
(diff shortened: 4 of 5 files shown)


src/Data/Scripts/Singletons/content_loader.gd
```diff
@@ -158,7 +158,7 @@ func _remove_top_most_mod_except_siblings(node: Node) -> void:
 			n.get_parent().remove_child(n)
 			dependency_tree_root.add_child(n)
 
-	top_most.free()
+	top_most.queue_free()
```
HaSa1002: This change is wrong. We can't queue free as the node is not inside the scene tree. We just use the nodes to build our hierachy :D

src/Data/Scripts/Rail.gd
```diff
@@ -174,9 +174,9 @@ func _update() -> void:
 func unload_visible_instance() -> void:
 	visible = false
 	if get_node_or_null("MultiMeshInstance") != null:
-		$MultiMeshInstance.free()
+		$MultiMeshInstance.queue_free()
 	if get_node_or_null("OverheadLine") != null:
-		$OverheadLine.free()
+		$OverheadLine.queue_free()
```
HaSa1002, line 179: those two changes feel safe. AFAIK these are only used to save performance

src/Data/Scripts/chunk_manager.gd
```diff
@@ -281,7 +281,7 @@ func _unload_old_chunks(all: bool = false):
 		var parent = world.get_node(group)
 		for child in parent.get_children():
 			if all or (child.get_meta("chunk_pos") in chunks_to_unload):
-				child.free()
+				child.queue_free()
```
HaSa1002: I am unsure about that one. @DasCapschen do you know more?

src/Editor/Docks/RailBuilder/RailBuilder.gd
```diff
@@ -175,7 +175,7 @@ func _on_Rename_pressed() -> void:
 
 
 func _on_Delete_pressed() -> void:
-	currentRail.free()
+	currentRail.queue_free()
```
HaSa1002, line 178: feels safe

src/Editor/Scripts/ScenarioConfiguration.gd
```diff
@@ -154,7 +154,7 @@ func update_ui_for_current_route():
 	$TabContainer/Routes/RouteConfiguration/GeneralSettings/P.visible = routes[current_route].general_settings.activate_only_at_specific_routes and not routes[current_route].general_settings.player_can_drive_this_route
 
 	for child in $TabContainer/Routes/RouteConfiguration/GeneralSettings/P/SpecificRoutes.get_children():
-		child.free()
+		child.queue_free()
```
HaSa1002, line 157: feels safe

src/Editor/Scripts/scenario_map.gd
```diff
@@ -173,9 +173,9 @@ func update_map():
 
 	# Clear old selection:
 	for child in $Special.get_children():
-		child.free()
+		child.queue_free()
 	for child in $RailsSelection.get_children():
-		child.free()
+		child.queue_free()
```
HaSa1002, line 178: feels safe

src/addons/jean28518.jTools/jTable/jTable.gd
```diff
@@ -48,10 +48,10 @@ func update_table_in_editor(newvar):
 func clear():
 	current_entries = 0
 	for child in grid_node.get_children():
-		child.free()
+		child.queue_free()
 
 	for heading in headings_node.get_children():
-		heading.free()
+		heading.queue_free()
```
HaSa1002, line 54: *should* be safe

HaSa1002 (changes requested):
Please fix the mentioned one.
I need @DasCapschen review, too.

src/Data/Scripts/chunk_manager.gd
DasCapschen: I think I used free() on purpose, because the chunks should unload immediately, but as I think queue_free() should work fine too.

DasCapschen (changes requested):
Looks good to me, apart from the one @HaSa1002 mentioned.

src/Data/Scripts/chunk_manager.gd
nalquas: > I think I used free() on purpose

In that case, I'll leave it as it is, so we'll keep the `free()` here.

HaSa1002 (approved):
lgtm

HaSa1002:
Thanks. Possible candidate for 0.9

PR #376, opened 2022-06-29
Input handling improvements

nalquas:
This pull request should improve some of the input handling code.

The changes are:

- In `Player.gd`, use the provided input events where appropriate, instead of unnecessarily checking `Input.is_action_pressed`.
- In `Player.gd`, move code from the custom `handle_input()` function to `_unhandled_input()`, thus making use of events and not actively checking for Input every frame.
- Somewhat modernize the movement code in `Camera.gd`. It now uses Vectors and Input Actions instead of manually checking the keyboard keys with if statements.

### DIFF
--- a/src/Data/Scripts/Camera.gd
+++ b/src/Data/Scripts/Camera.gd
@@ -107,17 +107,22 @@ func _process(delta: float) -> void:
 		translation.x -= missingCameraPosition * delta
 
 	if not fixed and (not Root.Editor or Input.is_mouse_button_pressed(BUTTON_RIGHT)):
-		var deltaFlyspeed: float = (delta / refDelta) * flyspeed
-
-		if(Input.is_key_pressed(KEY_W)):
-			self.set_translation(self.get_translation() - get_global_transform().basis*Vector3(0,0,1) * deltaFlyspeed)
-		if(Input.is_key_pressed(KEY_S)):
-			self.set_translation(self.get_translation() - get_global_transform().basis*Vector3(0,0,1) * -deltaFlyspeed)
-		if(Input.is_key_pressed(KEY_A) and not Input.is_key_pressed(KEY_CONTROL)):
-			self.set_translation(self.get_translation() - get_global_transform().basis*Vector3(1,0,0) * deltaFlyspeed)
-		if(Input.is_key_pressed(KEY_D)):
-			self.set_translation(self.get_translation() - get_global_transform().basis*Vector3(1,0,0) * -deltaFlyspeed)
-		if(Input.is_key_pressed(KEY_SHIFT)):
-			flyspeed = 2
+		# Handle the camera speed toggle
+		if(Input.is_action_pressed("shift")):
+			flyspeed = 2.0
 		else:
 			flyspeed = 0.5
+		
+		# Account for delta time
+		var deltaFlyspeed: float = (delta / refDelta) * flyspeed
+
+		# Only apply movement if CTRL isn't pressed
+		# (Workaround for CTRL+A "Autopilot" overlapping with A "move left")
+		if not Input.is_key_pressed(KEY_CONTROL):
+			# Get current analog input direction
+			var direction_2d := Input.get_vector("left", "right", "forward", "backward")
+			var direction := Vector3(direction_2d[0], 0.0, direction_2d[1])
+			
+			# Apply the movement
+			if direction.length() != 0.0:
+				self.translate(direction * deltaFlyspeed)
--- a/src/Data/Scripts/Player.gd
+++ b/src/Data/Scripts/Player.gd
@@ -338,8 +338,6 @@ func _process(delta: float):
 
 	check_station(delta)
 
-	handle_input()
-
 	currentRailRadius = currentRail.radius
 
 	if not ai:
@@ -351,7 +349,7 @@ func _process(delta: float):
 
 
 func _unhandled_key_input(event) -> void:
-	if event.is_action_pressed("debug") and not ai:
+	if event.is_action_pressed("debug") and not event.is_echo() and not ai:
 		debug = !debug
 		if debug:
 			send_message("DEBUG_MODE_ENABLED")
@@ -401,9 +399,28 @@ func _unhandled_input(event) -> void:
 		return
 	if event is InputEventMouseMotion:
 		mouseMotion = mouseMotion + event.relative
+	
+	# The following input events should only happen when the button is first pressed,
+	# not when it is held down. Hence, we'll return early if it's an echo.
+	if event.is_echo():
+		return
 
 	if event.is_action_pressed("acc+") and reverser == ReverserState.NEUTRAL:
 		send_message("HINT_REVERSER_NEUTRAL", ["reverser+", "reverser-"])
+
+	if event.is_action_pressed("InsideLight"):
+		toggle_cabin_light()
+	elif event.is_action_pressed("FrontLight"):
+		toggle_front_light()
+
+	if event.is_action_pressed("Horn"):
+		horn()
+
+	if event.is_action_pressed("reverser+"):
+		change_reverser(+1)
+
+	if event.is_action_pressed("reverser-"):
+		change_reverser(-1)
 
 	# Handle camera zoom
 	if event.is_action_pressed("zoom_in") or event.is_action_pressed("zoom_out"):
@@ -1721,25 +1738,6 @@ func change_reverser(change: int) -> void:
 				jAudioManager.play_game_sound("res://Resources/Sounds/click.ogg")
 
 	emit_signal("reverser_changed", reverser)
-
-
-func handle_input() -> void:
-	if ai:
-		return
-	if Input.is_action_just_pressed("FrontLight") and not Input.is_key_pressed(KEY_CONTROL):
-		toggle_front_light()
-
-	if Input.is_action_just_pressed("InsideLight"):
-		toggle_cabin_light()
-
-	if Input.is_action_just_pressed("Horn"):
-		horn()
-
-	if Input.is_action_just_pressed("reverser+"):
-		change_reverser(+1)
-
-	if Input.is_action_just_pressed("reverser-"):
-		change_reverser(-1)
 
 
 func fail_scenario(text: String) -> void:
--- a/src/Data/Scripts/Player.gd
+++ b/src/Data/Scripts/Player.gd
@@ -350,8 +350,8 @@ func _process(delta: float):
 	check_overdriving_a_switch()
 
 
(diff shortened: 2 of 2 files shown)


src/Data/Scripts/Camera.gd
```diff
@@ -107,17 +107,25 @@ func _process(delta: float) -> void:
...
+
+		# Workaround for CTRL+A (Autopilot) overlapping with A (move left)
+		var left_strength = Input.get_action_strength("left") if not Input.is_key_pressed(KEY_CONTROL) else 0.0
+
+		# Get current (analog) input direction
+		var direction = Vector3(\
+		Input.get_action_strength("right") - left_strength, \
+		0.0, \
+		Input.get_action_strength("backward") - Input.get_action_strength("forward")\
+		).normalized()
```
HaSa1002: FYI: https://docs.godotengine.org/en/stable/tutorials/inputs/controllers_gamepads_joysticks.html?highlight=Input.get_vector#which-input-singleton-method-should-i-use

src/Data/Scripts/Player.gd
```diff
@@ -405,6 +403,20 @@ func _unhandled_input(event) -> void:
...
+
+	if event.is_action_pressed("Horn"):
+		horn()
+
+	if event.is_action_pressed("reverser+"):
+		change_reverser(+1)
+
+	if event.is_action_pressed("reverser-"):
+		change_reverser(-1)
+
```
HaSa1002, line 424: The reason we used the input filter (and to be honest there is no shame in doing so) is that it supplies **is_action_just** methods. IMHO using the Input singleton instead of events is not an issue

HaSa1002 (changes requested):
I don't think this change really improves the code, tbh. Except for the rewrite of the camera input code. I do think @DasCapschen should take a look though.

HaSa1002:
Hmm. The docs do encourage a different style: https://docs.godotengine.org/en/stable/tutorials/inputs/input_examples.html?highlight=Input So I guess these changes may be fine except for the non existent echoing check. So you'll need to return early `if event.is_echo()`

src/Data/Scripts/Camera.gd
DasCapschen: can't we just do something like
`if  not Input.is_key_pressed(KEY_CONTROL): var direction = Input.get_vector(...)` ?

src/Data/Scripts/Player.gd
```diff
@@ -350,8 +348,8 @@ func _process(delta: float):
 	check_overdriving_a_switch()
 
 
-func _unhandled_key_input(_event) -> void:
-	if Input.is_action_just_pressed("debug") and not ai:
+func _unhandled_key_input(event) -> void:
+	if event.is_action_pressed("debug") and not ai:
```
DasCapschen: Does `is_action_pressed` actually behave the same as `is_action_*just*_pressed`? Or does `is_action_pressed` return true as long as the action is held down?

If it does not behave the same, please stick with `is_action_just_pressed`.

src/Data/Scripts/Player.gd
```diff
@@ -402,26 +400,39 @@ func _unhandled_input(event) -> void:
 	if event is InputEventMouseMotion:
 		mouseMotion = mouseMotion + event.relative
 
-	if Input.is_action_just_pressed("acc+") and reverser == ReverserState.NEUTRAL:
+	if event.is_action_pressed("acc+") and reverser == ReverserState.NEUTRAL:
```
DasCapschen, line 408: same as above

src/Data/Scripts/Player.gd
```diff
@@ -1724,25 +1735,6 @@ func change_reverser(change: int) -> void:
 	emit_signal("reverser_changed", reverser)
 
 
-func handle_input() -> void:
```
DasCapschen, line 1727: this is a good idea, thanks 😄

DasCapschen (changes requested):
Please check the comments.

src/Data/Scripts/Player.gd
nalquas: As I understand it:
Since we're using the InputEvent instead of Input, `event.is_action_pressed` should only trigger when the action is first activated (like with `Input.is_action_just_pressed`). However, holding down a key would result in many presses happening (like when you're holding down a key in a text editor). We can prevent this by ensuring that `not event.is_echo()` is true, though.

src/Data/Scripts/Camera.gd
nalquas: > can't we just do something like `if not Input.is_key_pressed(KEY_CONTROL): var direction = Input.get_vector(...)` ?

Good idea. You wouldn't want to move while holding down CTRL anyway, so this should work pretty well / looks more clean.

nalquas:
I've further improved the Camera.gd movement rewrite, which now uses `Input.get_axis` instead of `Input.get_action_strength`.
As for the `event.is_action_pressed` vs `Input.is_action_just_pressed` question, I've added checks for `event.is_echo()` to ensure that it behaves the same.

src/Data/Scripts/Camera.gd
```diff
@@ -107,17 +107,25 @@ func _process(delta: float) -> void:
...
+
+		# Only apply movement if CTRL isn't pressed
+		# (Workaround for CTRL+A "Autopilot" overlapping with A "move left")
+		if not Input.is_key_pressed(KEY_CONTROL):
+			# Get current (analog) input direction
+			var direction = Vector3(\
+				Input.get_axis("left", "right"), \
+				0.0, \
+				Input.get_axis("forward", "backward") \
+			)
```
HaSa1002:
```suggestion
			var direction := Vector3(\
				Input.get_axis("left", "right"), \
				0.0, \
				Input.get_axis("forward", "backward") \
			).normalized()
```
Please use typed variables where possible. 

FYI, not doing `.normalized()` enables players to fly faster, when going directional.

HaSa1002 (changes requested):

src/Data/Scripts/Camera.gd
nalquas: > Please use typed variables where possible.

I didn't realize static typing was a thing in GDScript. Makes sense to use it though.
 
> FYI, not doing `.normalized()` enables players to fly faster, when going directional.

I think an advantage of *not* using `.normalized()` would be future support for analog input (e.g. analog sticks / joysticks), however I see your point. I'll re-add it.

HaSa1002: > I didn't realize static typing was a thing in GDScript. Makes sense to use it though.

👍 

> I think an advantage of not using .normalized() would be future support for analog input (e.g. analog sticks / joysticks), however I see your point. I'll re-add it.

Input,get_vector has support for that built-in. It also has a circular deadzone instead of a squared one. And btw. we already inexplicitely support controllers. Have a look at the key bindings :D

nalquas: Hm, then I'll try to use `get_vector` instead. I didn't use it because it only gives us a `Vector2` and I didn't want to manually have to add another dimension to get a `Vector3`, but it's pretty trivial nonetheless.

src/Data/Scripts/Camera.gd
```diff
@@ -107,17 +107,22 @@ func _process(delta: float) -> void:
...
+		# Only apply movement if CTRL isn't pressed
+		# (Workaround for CTRL+A "Autopilot" overlapping with A "move left")
+		if not Input.is_key_pressed(KEY_CONTROL):
+			# Get current analog input direction
+			var direction_2d := Input.get_vector("left", "right", "forward", "backward")
+			var direction := Vector3(direction_2d[0], 0.0, direction_2d[1])
+			
+			# Apply the movement
+			if direction.length() != 0.0:
+				self.translate(direction * deltaFlyspeed)
```
DasCapschen, line 128: One more thing I've just noticed here: 
did you actually test how the camera moves now? 

Because `direction` now only depends on the input, but not on the current rotation of the camera.
While before, `get_global_transform().basis` was used to account for the current rotation of the camera.

Meaning, when pressing forward, you now wouldn't actually move forward, but along the Z axis.
Unless I've missed something. 🤔 
I believe `var direction := Vector3(...) * get_global_transform().basis` should do it.

What do you think @HaSa1002 

(unfortunately, I don't have the time to test it myself right now)

src/Data/Scripts/Player.gd
```diff
@@ -401,27 +399,45 @@ func _unhandled_input(event) -> void:
...
+	if event.is_action_pressed("reverser+"):
+		change_reverser(+1)
+
+	if event.is_action_pressed("reverser-"):
+		change_reverser(-1)
+
+	# Handle camera zoom
+	if event.is_action_pressed("zoom_in") or event.is_action_pressed("zoom_out"):
+		# zoom out
+		if event.is_action_pressed("zoom_out"):
```
DasCapschen, line 428: Why are those ifs nested? 🤔 
I feel like we don't need `if event.is_action_pressed("zoom_in") or event.is_action_pressed("zoom_out"):`

DasCapschen (changes requested):

src/Data/Scripts/Camera.gd
nalquas: The rotation is accounted for by `self.translate(...)`. It seems to work just fine for me:

https://user-images.githubusercontent.com/22664233/177387462-dfd48f68-0503-48e6-9515-f25830398054.mp4

DasCapschen: Okay, thank you for confirming that. :) 
Sorry 😅

src/Data/Scripts/Player.gd
nalquas: Because lines 442-443 (which are in that if clause's block) only really need to happen when one of the zoom events happened.
https://github.com/Libre-TrainSim/Libre-TrainSim/blob/9d4bafcff6179e3bc17fc11a5b9924b44c4b8e62/src/Data/Scripts/Player.gd#L442-L443

Although before, it was just always called if `event.is_pressed()` was true and a comment in the zoom blocks stated to "call the zoom function" (which doesn't exist, therefore I assume those two lines were meant).

DasCapschen: True, I've missed that, sorry.
Everything looks fine then :)

DasCapschen (approved):

HaSa1002 (changes requested):
lgtm. Please clean up the history

nalquas:
> lgtm. Please clean up the history

Done.

HaSa1002 (approved):

HaSa1002:
Thank you very much ❤️

PR #380, opened 2022-07-23
Add FPS Limiter to Settings (requires Godot 3.5)

nalquas:
This PR adds an FPS limiter to the settings.

The FPS limiter is useful when VSync is disabled. It should allow users of high-end hardware to limit FPS to their monitor refresh rate to make use of GSync/FreeSync. It would also allow laptop users to limit their game to, for example, 30fps for reduced power consumption.

I marked this as a draft since this requires Godot 3.5's [OS.get_screen_refresh_rate()](https://docs.godotengine.org/en/3.5/classes/class_os.html?highlight=OS#class-os-method-get-screen-refresh-rate) (see https://github.com/godotengine/godot-proposals/issues/1284), but Godot 3.5 hasn't been released yet. We should revisit this when 3.5 is out.

To be fair, OS.get_screen_refresh_rate() isn't entirely necessary and we could just initialize the FPS limit to some default value (e.g. 60FPS), but the user experience is better this way.

### DIFF
--- a/src/Translations/MainMenu.csv
+++ b/src/Translations/MainMenu.csv
@@ -26,6 +26,8 @@ MENU_MUSIC:,Music: Follow That Dream by Luca Fraula,Musik: Follow That Dream von
 MENU_SETTINGS:,Settings:,Einstellungen:,Impostazioni:,設定:,Ustawienia,Налаштування:,Configuración:,Настройки:,Paramètres :
 MENU_FULLSCREEN,Fullscreen,Vollbild,Schermo intero,フルスクリーン,Pełny ekran,Повноекранний режим,Pantalla completa,На цял екран,Plein écran
 MENU_VSYNC,Enable Vertical Synchronisation,Vertikale Synchronisation aktivieren,,,,,,,
+MENU_FPS_LIMIT_TOGGLE,Enable FPS Limit,FPS-Limit aktivieren,,,,,,,
+MENU_FPS_LIMIT,FPS Limit,FPS-Limit,,,,,,,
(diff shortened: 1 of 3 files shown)


nalquas:
Since Godot 3.5 [is now officially released](https://godotengine.org/article/godot-3-5-cant-stop-wont-stop), I'll mark this as ready for review. I've cleaned up the proposed code a bit, too.

This PR is now using the following Godot 3.5 features:
- `OS.get_screen_refresh_rate()`
- The new "Access nodes by unique names" feature, allowing use of `$"%unique_node_name"` instead of `$very/long/path/to/the/node/we/actually/want`

The GUI will also dynamically show/hide the FPS limiter settings depending on whether VSync is enabled or not.

HaSa1002:
I wanna check it out locally but otherwise it looks good to me. I am a little bit torn by the the unique names because technically they are a get_node every time we access these. But that is something we may wanna decide for more perf critical code. It's completely fine here.

This is blocked until we have CI updated for 3.5 (waiting for a new release: https://github.com/abarichello/godot-ci)

HaSa1002:
Also on that note, the fact CI passes makes me wonder if we can actually manage to let CI fail with wrong code. Would be interesting to have some more safety checks

HaSa1002 (dismissed):
lgtm. Working as intended. My graphics card put on the lovely sound of coil whine when I disabled VSync and disabled it again when I enabled the FPS limiter.

Only waiting for the CI update

HaSa1002:
A rebase is required.

nalquas:
> A rebase is required.

Done

HaSa1002 (approved):
lgtm

HaSa1002:
Thanks! ❤️

PR #381, opened 2022-07-31
Refactor configuration into window

### LINKED ISSUE #382: Editor details menus opens themselves even if they've been hidden by the user
The editor details menu is shown every time you click an object which is quite annoying tbh. Why have a show/hide settings button then anyway?

HaSa1002:
I also made some adjustments to the buttons in the top left (from now on global menu). Most importantly making the jump station button a MenuButton and putting all of these buttons into a container.

The config window is instantiated every time and saves its values when pressing save. Cancel, close, or pressing escape won't save the values.
Speaking of escape... In order to fix escape really working the way we wanna have it working, I had to refactor the menu into its own script. So that's that. I took the chance and replaced `get_parent().get_parent().save_world()` with a signal, too.

Sorry for that huge boy of a commit. I am out of practice :sweat_smile:

Fixes #382

### DIFF
--- a/src/Editor/Scripts/pause_menu.gd
+++ b/src/Editor/Scripts/pause_menu.gd
@@ -1,0 +1,30 @@
+extends Control
+
+signal save_requested
+
+
+func _unhandled_key_input(event: InputEventKey) -> void:
+	if visible and Input.is_action_just_released("ui_cancel", true):
+		# There seems to be an issue with event propagation
+		# Hence we wait for the frame to end before we actually hide it
+		# because otherwise we immediately open the menu again.
+		call_deferred("hide")
+
+
+func _on_SaveAndExit_pressed():
+	emit_signal("save_requested")
+	get_tree().paused = false
+	LoadingScreen.load_main_menu()
+
+
+func _on_Back_pressed() -> void:
+	hide()
+
+
+func _on_hide() -> void:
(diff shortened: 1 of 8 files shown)


src/Editor/Scripts/pause_menu.gd
```diff
@@ -0,0 +1,30 @@
...
+
+func _on_SaveAndExit_pressed():
+	emit_signal("save_requested")
+	get_tree().paused = false
+	LoadingScreen.load_main_menu()
+
+
+func _on_SaveWithoutExit_pressed():
+	get_tree().paused = false
+	LoadingScreen.load_main_menu()
```
DasCapschen: This isn't `SaveWithoutExit`, this is `ExitWithoutSave`

DasCapschen (changes requested):
Haven't tried it, but the code mostly looks good. 
Thank you.

src/Editor/Scripts/pause_menu.gd
HaSa1002: uh lol. I just copied the code and that is the name Jean gave it xD

HaSa1002: fixed

DasCapschen (approved):
lgtm

PR #390, opened 2022-08-20
Feature: TSCN Chunks

### LINKED ISSUE #150: Use scenes and resources to store chunks and scenarios

### LINKED ISSUE #388: Track Editor: Saving makes all tracks invisible
### Summary
In the track editor, pressing "Save World" or using `Ctrl+S` makes all tracks go invisible.

### How to reproduce
1. Open any track in the track editor
2. Save (Using "Save World" or `Ctrl+S`)

### Observed behavior
All tracks go invisible, although the direction indicators (the red/green arrows and the white one in the middle) stay visible.
Clicking one of these indicators makes its track piece appear again.

### Expected behavior
The world is saved and the tracks stay visible.

### Version
 - Game Version: commit 64586c1c80c69a66a153bbc886966ca34dea00a5
 - OS: Linux

### Additional information
Returning to menu and opening the track again makes the tracks go visible again.
Bug appeared after https://github.com/Libre-TrainSim/Libre-TrainSim/pull/277 was merged.

DasCapschen:
Replaces PR #236  
I did not want to rebase that... it had a lot of refactoring in it that wasn't related to TSCN chunks.  
So I rewrote it. This now also supports the new Scenario System.

THIS IS CURRENTLY UNTESTED. But all the necessary code changes should be there, apart from possibly bug fixes 😄 
Will test and fix tomorrow, hopefully 😅 

Completely replaces the `.save`, `.trackinfo` and `.scenario` files with `.tscn` files.
`jSaveModule` is no longer used anywhere.

Worlds have not yet been converted!

fixes #388 
fixes #150


src/Data/Scripts/DataClasses/rail_logic_settings.gd
```diff
@@ -0,0 +1,2 @@
+class_name RailLogicSettings
+extends Reference
```
HaSa1002: Aren't those meant to be serialised? If so, it'd make sense to extend from Resource and `export` the serialisable values

src/Data/Scripts/DataClasses/contact_point_settings.gd
```diff
@@ -0,0 +1,10 @@
+class_name ContactPointSettings
+extends RailLogicSettings
+
+var enabled := false
+var affected_signal := ""
+var affect_time := 0.1
+var new_speed_limit := -1
+var new_status := 1
+var enable_for_all_trains := true
+var specific_train := ""
```
HaSa1002: Are these meant to be serialised?

src/Data/Scripts/DataClasses/route_point.gd
```diff
@@ -0,0 +1,8 @@
+class_name RoutePoint
+extends Reference
+
+var type := RoutePointType.NONE
+
+func get_description() -> String:
+	return "Unknown"
+
```
HaSa1002: Same Resource question

src/Data/Scripts/DataClasses/route_point_despawn.gd
```diff
@@ -0,0 +1,12 @@
+class_name RoutePointDespawnPoint
+extends RoutePoint
+
+var rail_name := ""
+var distance_on_rail := 0.0
+
+func _init() -> void:
+	type = RoutePointType.DESPAWN_POINT
```
HaSa1002: :| C'mon GDScript. That's ugly
But actually you could just do type checks on the object itself.

src/Data/Scripts/DataClasses/scenario_route.gd
```diff
@@ -0,0 +1,214 @@
+class_name ScenarioRoute
+extends Reference
```
HaSa1002: Again, `Resource`?

src/Data/Scripts/DataClasses/scenario_route.gd
```diff
@@ -0,0 +1,214 @@
...
+
+func get_calculated_rail_route(world: Node) -> Array:
+	world.update_rail_connections() # why is this necessary? :(
+	var rail_route: Array = []
+
+	for i in range (size()-1):
+		# get start and end rails for calculation
+		var start_end_rails: Array = []
+		var start_direction_set: bool = false
+		var forward: bool = true
```
HaSa1002: You know, type inference is a thing? xD No need to be overly verbose when not using number

src/Data/Scripts/DataClasses/scenario_route.gd
```diff
@@ -0,0 +1,214 @@
...
+
+		# wtf is this for loop even doing here???
+		for j in range(2):
+			var route_point: Dictionary = route_points[i+j]
+			if route_point.type == RoutePointType.STATION:
```
HaSa1002:
```suggestion
			if route_point is RoutePointStation:
```

src/Data/Scripts/DataClasses/scenario_route.gd
```diff
@@ -0,0 +1,214 @@
...
+	return rail_route
+
+
+# call get_calculated_rail_route() before!
+func get_spawn_point(train_length, world: Node) -> RoutePointSpawnPoint:
+	var start_point: RoutePoint = route_points[0]
+
+	if start_point is RoutePointSpawnPoint:
+		start_point.forward = calculated_rail_route[0].forward
+		return start_point as RoutePointSpawnPoint
```
HaSa1002: No need for the cast. We are duck typed.

src/Data/Scripts/DataClasses/scenario_route.gd
```diff
@@ -0,0 +1,214 @@
...
+		spawn_point.forward = station_node.forward
+		spawn_point.initial_speed = 0
+		spawn_point.initial_speed_limit = -1
+		return spawn_point
+
+	return null
+
+
+func get_despawn_point() -> RoutePointDespawnPoint:
+	return route_points.back()  # cannot duplicate() objects :(
```
HaSa1002: You could duplicate `Resource`s

src/Data/Scripts/DataClasses/train_spawn_information.gd
```diff
@@ -0,0 +1,11 @@
+class_name TrainSpawnInformation
+extends Reference
```
HaSa1002: Same `Resource` question.

src/Data/Scripts/Enums/route_point_type.gd
```diff
@@ -1,6 +1,7 @@
 class_name RoutePointType
 
 enum {
+	NONE = -1,
 	STATION = 0,
 	WAY_POINT = 1,
 	SPAWN_POINT = 2,
```
HaSa1002: Using types we can get rid of them

src/Data/Scripts/Player.gd
```diff
@@ -1550,18 +1550,19 @@ func checkDespawn() -> void:
 	if not ai:
 		return
 
-	if despawn_information.type == RoutePointType.STATION and despawn_information.stop_type == StopType.END:
+	if despawn_point.type == RoutePointType.STATION and despawn_point.stop_type == StopType.END:
```
```suggestion
	if despawn_point is RoutePointStation and despawn_point.stop_type == StopType.END:
```

src/Data/Scripts/Player.gd
```diff
@@ -1550,18 +1550,19 @@ func checkDespawn() -> void:
...
 		# Despawn if the train has arrived at the endstation and the planned arrival is two minutes in the past
-		if world.time > despawn_information.arrival_time + 60*2 and get_station_table_index_of_station_node_name(despawn_information.node_name) >= current_station_table_index:
+		if world.time > despawn_point.arrival_time + 60*2 and get_station_table_index_of_station_node_name(despawn_point.node_name) >= current_station_table_index:
 			despawn()
 
-	if despawn_information.type == RoutePointType.DESPAWN_POINT:
-		if currentRail.name == despawn_information.rail_name:
-			if forward and distance_on_rail > despawn_information.distance:
+	if despawn_point.type == RoutePointType.DESPAWN_POINT:
```
```suggestion
	if despawn_point is RoutePointDespawn:
```

src/Data/Scripts/Resources/track_scenario.gd
```diff
@@ -0,0 +1,44 @@
...
+export var time: int = 0  # seconds
+export var title := ""
+export var description := ""
+export var duration: int = 0  # minutes
+
+export var is_hidden := false  # true = visible only in Editor, not in Play menu
+
+# Dict[String, ScenarioRoute]
+# see World.gd spawn_train()
+export var routes := {}
```
HaSa1002: Can we set the types or is that a 4.0 reserved feature?
```suggestion
export(String, ScenarioRoute) var routes := {}
```

src/Data/Scripts/Resources/track_scenario.gd
```diff
@@ -0,0 +1,44 @@
...
+
+# Dict[String, RailLogicSettings]
+# see Signal.gd get_scenario_data()
+export var rail_logic_settings := {}
```
HaSa1002: see above

src/Data/Scripts/Resources/track_scenario.gd
```diff
@@ -0,0 +1,44 @@
...
+	return current_scenario
+
+
+func save_scenario(path = null):
+	if path == null:
+		path = Root.current_track.get_base_dir().plus_file("scenarios").plus_file(Root.current_scenario) + ".tres"
+
+	if ResourceSaver.save(path, self) != OK:
+		Logger.err("Failed saving scenario at %s" % path, self)
+	pass
```
HaSa1002:
```suggestion
```

src/Data/Scripts/TrackObjects.gd
```diff
@@ -110,12 +106,18 @@ func _exit_tree() -> void:
 
 
 func _ready() -> void:
-	Root.connect("world_origin_shifted", self, "_on_world_origin_shifted")
```
HaSa1002, line 113: Was that a bug?

src/Data/Scripts/World.gd
```diff
@@ -8,7 +8,7 @@ var default_persons_at_station: int = 20
 
 var globalDict := {} ## Used, if some nodes need to communicate globally. Modders could use it. Please make sure, that you pick an unique key_name
```
HaSa1002, line 9: Can we please delete that?

src/Data/Scripts/World.gd
```diff
@@ -140,69 +117,61 @@ func get_signal_scenario_data() -> Dictionary:
...
-	var route_name = ""
-	var minimal_platform_length = 0
-	var route = []
-	var station_table = []
-	var despawn_information = {}
-
-
 func set_scenario_to_world() -> void:
-	scenario_manager.set_save_path(Root.current_scenario)
+	current_scenario = TrackScenario.load_scenario()
```
HaSa1002, line 127: I possibly forgot but how do we determine the scenario?

src/Data/Scripts/chunk.gd
```diff
@@ -0,0 +1,52 @@
+class_name Chunk
+extends Spatial
+
+export var is_empty := true
+export var generate_grass := true  # set this to false when we have terrain
```
HaSa1002, line 5: Can we add support for abritary image loading?
We kinda need this as an editor feature and Jean's documentation builds around "Importing a base image"
https://docs.libretrainsim.org/en/merge-pages/manual/editor/01-first-steps/#conception-phase
https://docs.libretrainsim.org/en/merge-pages/manual/editor/02-getting-started/#base-image

src/Data/Scripts/chunk.gd
```diff
@@ -0,0 +1,52 @@
...
+# array of node names
+export (Array, String) var rails := []
+
+
+# see chunk_prefab.tscn
+# TrackObjects are saved in global coordinates, NOT relative to the chunk!
+# this is because they are attached to the Rails!
+# Buildings are also stored in global coordinates
+# DefaultGrass is stored relative to the chunk, we must move it to the correct global position
+# thus, Chunks should always be at (0,0,0)
```
HaSa1002, line 19: I am not overly happy with that conclusion. The global coordinates should be made local to avoid floating point issues with the serialisation and then mismatch when we move and load the chunks.

src/Data/Scripts/chunk.gd
```diff
@@ -0,0 +1,52 @@
...
+func _on_world_origin_shifted(delta):
+	if generate_grass:
+		$DefaultGrass.translation += delta
+
+	# FIXME: this does not actually work...
+	#        it would be so nice, if we could just shift the entire chunk node
+	#        instead of all the track objects,
+	#        but I think that will require a TrackObject refactor
+	#        ask me about it for 0.10
+	#translation += delta
```
HaSa1002, line 81: I am here to ask you @DasCapschen :P

src/Data/UI/play_menu.gd
```diff
@@ -126,13 +124,13 @@ func update_trains():
 func _on_Tracklist_item_selected(index) -> void:
 	selected_track = ContentLoader.repo.worlds[index]
 	Root.checkAndLoadTranslationsForTrack(selected_track.get_file().get_basename())
-	var save_path: String = ContentLoader.repo.worlds[index].get_basename() + ".trackinfo"
-	j_save_module.set_save_path(save_path)
+	var config_path: String = ContentLoader.repo.worlds[index].get_basename() + "_config.tres"
+	world_config = load(config_path) as WorldConfig
```
HaSa1002: No need to cast
```suggestion
	world_config = load(config_path)
```

src/Editor/Docks/Configuration/Configuration.gd
```diff
@@ -1,13 +1,11 @@
...
-var j_save_module := jSaveModule.new()
-
+var world_config: WorldConfig
+var save_path: String
 
 func _ready() -> void:
-	save_path = find_parent("Editor").current_track_path + ".trackinfo"
-	j_save_module.set_save_path(save_path)
+	save_path = find_parent("Editor").current_track_path + "_config.tres"
+	world_config = load(save_path) as WorldConfig
```
HaSa1002: No need to cast.
```suggestion
	world_config = load(save_path)
```

src/Editor/Scripts/ScenarioConfiguration.gd
```diff
@@ -78,58 +68,39 @@ func show_selection_message(text: String) -> void:
...
 
 
 func _on_Routes_user_duplicated_entries(source_entry_names, duplicated_entry_names):
 	var source_entry_name = source_entry_names[0]
 	var duplicated_entry_name = duplicated_entry_names[0]
-	routes[duplicated_entry_name] = routes[source_entry_name].duplicate(true)
-	j_save_module.save_value("routes", routes)
-	j_save_module.reload()
-	routes = j_save_module.get_value("routes", {})
+	routes[duplicated_entry_name] = routes[source_entry_name]
```
HaSa1002: Isn't that a reference?

src/Editor/Scripts/ScenarioConfiguration.gd
```diff
@@ -272,7 +244,7 @@ func update_route_point_settings():
 		return
 	var index = selected_items[0]
 
-	var route_point = route_manager.get_point(index)
+	var route_point = loaded_route.get_point(index)
 	if route_point.type == RoutePointType.STATION:
```
HaSa1002:
```suggestion
	if route_pointis RoutePoinStation:
```

HaSa1002: Probably more down here

src/Editor/Scripts/ScenarioConfiguration.gd
```diff
@@ -801,18 +780,18 @@ func _on_CheckIfRouteIsValid_pressed():
...
 		return
 
-	if route_manager.get_route_data().size() < 2:
+	if route_points.size() < 2:
 		error_message += "Your route has to have at least two route points!\n\n"
 		$TabContainer/Routes/RouteConfiguration/IsRouteValid/Messages.text = error_message
 		return
 
-	var station_points = route_manager.get_calculated_station_points(0)
+	var station_points = loaded_route.get_calculated_station_points(0)
```
HaSa1002, line 797: A ton of route point types are following

src/project.godot
```diff
@@ -44,6 +49,11 @@ _global_script_classes=[ {
 "language": "GDScript",
 "path": "res://Data/Scripts/ContactPoint.gd"
 }, {
+"base": "RailLogicSettings",
+"class": "ContactPointSettings",
+"language": "GDScript",
+"path": "res://Data/Scripts/DataClasses/contact_point_settings.gd"
```
HaSa1002: Please rename `DataClasses` to underscore_case

HaSa1002 (changes requested):
A tired review. Mostly questions

src/Data/Scripts/DataClasses/rail_logic_settings.gd
DasCapschen: Uh, I haven't done this in a while. Can Reference not be serialised? In that case, you're right, those need to be Resource and export their variables. I'll test that later today.

src/Data/Scripts/DataClasses/route_point_despawn.gd
DasCapschen: You're correct, this Type Enum is not really needed anymore, since I refactored it into classes 😅

src/Data/Scripts/DataClasses/scenario_route.gd
DasCapschen: This is just Jeans code that I copied over :D 
I replaced the ScenarioManager and RouteManager with simply the scenario and the route.

src/Data/Scripts/DataClasses/scenario_route.gd
DasCapschen: Okay, you've convinced me, I'll make them all Resources :smile:

src/Data/Scripts/Resources/track_scenario.gd
DasCapschen: I think this didn't work, because "ScenarioRoute is not a ResourceType". I believe this happens even if it `extends Resource`, only built-in types can be exported like that.

src/Data/Scripts/World.gd
DasCapschen: The player selects it in the play menu, which sets `Root.current_scenario`.
`TrackScenario.load_scenario()` loads `Root.current_scenario` by default, if not passed a path to a scenario file.

src/Data/Scripts/World.gd
DasCapschen: Yes, but can we do it in a different PR? 😅

src/Data/Scripts/chunk.gd
DasCapschen: See the next note, it requires a TrackObject refactor, afaik. I agree, but could we do it in a different PR?  
The way chunks work now (before tscn) is also in global coordinates, because everything gets loaded in as an individual object.  
So let's try and make it work the old way in the new format, and then refactor it to make it better :)

src/Editor/Scripts/ScenarioConfiguration.gd
DasCapschen: You might be right there 🤔 
Will solve by using Resource type and .duplicate() :)

src/project.godot
DasCapschen: Are you sure? All our folders are CamelCase at the moment, only files use snake_case 🤔

src/Editor/Docks/Configuration/Configuration.gd
DasCapschen: Are you sure? It is loaded from a file, doesn't it need to know the type then? 🤔 
I see `as type` on a lot of `load(...)` calls.

src/Data/Scripts/DataClasses/scenario_route.gd
DasCapschen: Nope, actually necessary because the function returns RoutePointSpawnPoint. Without the cast, Godot complains that `the returned type (RoutePoint) doesn't match the functions return type (RoutePointSpawnPoint)` :)

DasCapschen:
I believe this feature is now finished.  
Feel free to review the code and more importantly, test it out in-game 😄 
@nalquas @HaSa1002 

(yes Hasa, I will rebase and cleanup the history after you've decided everything is good 😉)

src/Editor/Docks/Configuration/Configuration.gd
HaSa1002: yeah but only when we set the variable type in the same call

HaSa1002: the type is defined earlier in this case

src/project.godot
HaSa1002: I'd wish to change all of that but this is going to make backports very... painful

DasCapschen: okay, anyways, this folder exists no more. I've moved them to Resources now, because, well, they are resources now 😄

src/Data/Scripts/chunk.gd
DasCapschen: This does indeed sound like a good idea, but again, I think it is worth it's own PR. Do we already have an Issue for this feature?

nalquas:
Currently (at bb1395dee523dfeb3c3a40874f1c937f1b848300), opening any track in the track editor crashes the game with an `Assertion failed.` in `res://Data/Scripts/World.gd:46 at function: _ready`, as well as the following two errors in the Debugger:
```
E 0:00:27.686   load_interactive: Cannot open file 'user://editor/tutorial/Worlds/tutorial/tutorial_config.tres'.
  <C++ Error>   Condition "err != OK" is true. Returned: Ref<ResourceInteractiveLoader>()
  <C++ Source>  scene/resources/resource_format_text.cpp:1183 @ load_interactive()
  <Stack Trace> World.gd:45 @ _ready()
                Editor.gd:572 @ load_world()
                Editor.gd:23 @ _ready()
                loading_screen.gd:127 @ _add_to_tree()
                loading_screen.gd:104 @ _clean_up_and_switch()
                loading_screen.gd:74 @ _process()
```
```
E 0:00:27.686   _load: Failed loading resource: user://editor/tutorial/Worlds/tutorial/tutorial_config.tres. Make sure resources have been imported by opening the project in the editor at least once.
  <C++ Error>   Condition "found" is true. Returned: RES()
  <C++ Source>  core/io/resource_loader.cpp:270 @ _load()
  <Stack Trace> World.gd:45 @ _ready()
                Editor.gd:572 @ load_world()
                Editor.gd:23 @ _ready()
                loading_screen.gd:127 @ _add_to_tree()
                loading_screen.gd:104 @ _clean_up_and_switch()
                loading_screen.gd:74 @ _process()
```
Edit: And yes, it happens with other maps than the Tutorial too, e.g. with self-created maps from before and after this PR's changes.

Edit 2: Interestingly, the crash does *not* happen when I export the game and run that build. In the Godot Editor however, it crashes for me every time.

DasCapschen:
That's interesting... I don't have any crashes, neither in Godot nor in export 🤔 
I will investigate why this happens.

HaSa1002:
> Edit 2: Interestingly, the crash does not happen when I export the game and run that build. In the Godot Editor however, it crashes for me every time.

@nalquas it's an assert. It is not executed in exports. It's a sanity check which acts like a contract to allow us to program with minimal overhead under expectation certain requirements are met. Not meeting these requirements is a code writing time issue and thus a bug.

HaSa1002:
The question really is: Is that assert a code writing time failure. Especially given it follows a load from the file system that may fail.
It may be a conversion issue but can you please share the content of the folder `user://editor/tutorial/Worlds/tutorial/`?

DasCapschen:
I suspect it somehow failed to save the `tutorial_config.tres`. 
It should be created when you convert an old map or create a new map in the Editor map selection screen.
Can you check your logs for further messages?

nalquas:
> @nalquas it's an assert. It is not executed in exports. It's a sanity check which acts like a contract to allow us to program with minimal overhead under expectation certain requirements are met. Not meeting these requirements is a code writing time issue and thus a bug.

You're right, I forgot (probably due to the time 🤷). Sorry


> The question really is: Is that assert a code writing time failure. Especially given it follows a load from the file system that may fail. It may be a conversion issue but can you please share the content of the folder `user://editor/tutorial/Worlds/tutorial/`?

Here you go: [tutorial.zip](https://github.com/Libre-TrainSim/Libre-TrainSim/files/9414600/tutorial.zip)


> I suspect it somehow failed to save the `tutorial_config.tres`. It should be created when you convert an old map or create a new map in the Editor map selection screen. Can you check your logs for further messages?

There aren't any further errors shown, however lots of warnings. I'm not sure which ones could be relevant, though, as there's nothing standing out of the random warnings. I can see a lot of `... returns a value, but this value is never used` warnings with quite a few of them coming from `name_node_appropriate` calls by `Editor.gd` or `RailBuilder.gd`. There's also some unused return values for `open()`, `make_dir_recursive()`, `change_dir()`, `pack()` and `save()`, with the former three coming from `export_track.gd` and the latter two coming from `Editor.gd`.
Aside from that, two variable shadowing warnings for `start_pos` and `end_pos` at `Rail.gd:478`.
Additionally, of course, there are the obligatory `Skipping pack loading in editor build, because of ...` warnings.
And lots of other random warnings.

So nothing really obviously relevant in the debugger or the logs.
Edit: Here's the `godot.log`, although it seems normal aside from the mentioned error at the end: [godot.log](https://github.com/Libre-TrainSim/Libre-TrainSim/files/9414747/godot.log)

### Debugging
Further Edits:
Edit 2: I checked in the debugger and `current_world_config` is `null` for me at the point where the `assert` is called.
Edit 3: It seems to be `null` because "user://editor/tutorial/Worlds/tutorial/tutorial_config.tres" doesn't exist.
Edit 4: I think `tutorial_config.tres` is supposed to be created in `Editor.gd`'s `_port_to_new_trackinfo()`? However, by setting a breakpoint both there and at the assert of the crash, I can see that `_port_to_new_trackinfo()` isn't called (yet).
Edit 5: Okay, going through it with the debugger step-by-step, I can see that `World.gd`'s `_ready()`, which contains the assert, is called by `Editor.gd`'s `load_world()` during `add_child(world)` (line 572). That in turn has been called by `Editor.gd`'s `_ready()` function in line 23. The `_port_to_new_trackinfo()` function would be called in line 38, so after the other stuff, and thus hasn't been called yet by the time the assert happens. @DasCapschen, does this help?

So to sum it up: From what I can tell, you're first loading the world (`Editor.gd:23`) and thereby calling an assert (`World.gd:45`) that `tutorial_config.tres` exists, and only then start converting the map to the new format (`Editor.gd:38`) and creating `tutorial_config.tres`. The error has to be here.

DasCapschen:
This was exactly the problem, thank you. I just replicated it on my side too :)
I have fixed this by calling `_port_to_new_trackinfo()` before calling `load_world()`

Your tutorial world should now import correctly 👍🏻

nalquas:
> Your tutorial world should now import correctly 👍🏻

Thanks, that fixed it.

I have another crash with another (admittedly very empty and non-important) [test map](https://github.com/Libre-TrainSim/Libre-TrainSim/files/9419233/testonly.zip) (zipped before conversion) though at `Editor.gd:230`: `Unable to iterate on object of type 'Nil'.` A simple check whether the old rail logic settings are empty should fix that, I think.

nalquas:
There also appears to be something wrong with the scenario conversion. I had two lines with different routes (stations) in the Tutorial map, however both lines now contain the stations of both original lines (Note the multiple First Stations, for example):

Screenshot_20220824_223058

Simple enough to fix by hand of course, but still, thought I'd report it.

HaSa1002:
Fair, I mean we are going to have fun converting Hainfurt and Nuremberg 😅

DasCapschen:
I've added a null check, give it a try when you can :)

DasCapschen:
> Fair, I mean we are going to have fun converting Hainfurt and Nuremberg

U2 is is almost done 😉 
There's just a missing tunnel somewhere I need to reconstruct 🙈

DasCapschen:
> There also appears to be something wrong with the scenario conversion. I had two lines with different routes (stations) in the Tutorial map, however both lines now contain the stations of both original lines (Note the multiple First Stations, for example):

I am aware of this bug, but I have no idea why it would happen or how to fix it.
It's a completely new function call for each scenario, somehow memory isn't cleared. 😦 
I think it might be jSaveModule.

nalquas:
> I've added a null check, give it a try when you can :)

Fixed that spot, but now there's another crash of the same type due to there also being no routes in the map:
`Unable to iterate on object of type 'Nil'.` at `Editor.gd:197` (`_convert_scenario`)

A similar solution should work here, too.

DasCapschen:
I am glad you are testing things I didn't even think of. Here I am assuming all old data is valid 😂 
Added a null check.

nalquas:
> I am glad you are testing things I didn't even think of. Here I am assuming all old data is valid :joy: Added a null check.

Minor issue with that fix: `Parser Error: A non-void function must return a value.` at `Editor.gd:263`

DasCapschen:
weird, Godot didn't complain to me about that 🤔 
fixed it.

Also, thank you very much for your work and review 🙏🏻

nalquas:
> weird, Godot didn't complain to me about that :thinking: fixed it.

Thanks, that worked. All my files are loadable now.

> Also, thank you very much for your work and review 🙏🏻

Glad to help :)

nalquas:
~~Actually, something else seems to be broken now. All of the objects in the map, aside from the rails, seem to have been moved into the middle of the map. Probably chunk-related? Maybe everything has been moved into the center chunk or something~~
<details>

Screenshot_20220824_234700

Here's the files for analysis:
[tutorial_before.zip](https://github.com/Libre-TrainSim/Libre-TrainSim/files/9419653/tutorial_before.zip)
[tutorial_after_conversion.zip](https://github.com/Libre-TrainSim/Libre-TrainSim/files/9419624/tutorial_after_conversion.zip)

</details>

Edit: Hm, tested it again and now everything works. No idea what happened. Issue appears to have been on my end this time

Thanks again for the fixes! :+1:

src/Data/Scripts/chunk.gd
nalquas: > This does indeed sound like a good idea, but again, I think it is worth it's own PR. Do we already have an Issue for this feature?

I've created an issue: https://github.com/Libre-TrainSim/Libre-TrainSim/issues/399

src/Data/Scripts/train_spawn_information.gd
```diff
@@ -0,0 +1,11 @@
...
+extends Reference
+
+var time := 0
+var player_train := false
+var train_path := ""
+var route_name := ""
+var minimal_platform_length := 0
+var route := []
+var station_table := []
+var despawn_point
```
nalquas:
```suggestion
var despawn_point: RoutePointDespawnPoint
```
We could type this, I think

DasCapschen: I think we can type it as `RoutePoint`, but we cannot type it as `RoutePointDespawnPoint`, because the despawn point can be either `RoutePointDespawnPoint` or `RoutePointStation`

DasCapschen:
> ~Actually, something else seems to be broken now. All of the objects in the map, aside from the rails, seem to have been moved into the middle of the map. Probably chunk-related? Maybe everything has been moved into the center chunk or something~
> 
> Edit: Hm, tested it again and now everything works. No idea what happened. Issue appears to have been on my end this time
> 
> Thanks again for the fixes! +1

Weird that you saw that issue, 0c73eaa9c68bef03b2b3b69f7382c8d23bae7f28 should have fixed that 🤔 
Good to know it worked when you tested it a second time though. Maybe it's inconsistent and only breaks some times? 🤔

src/Data/Scripts/train_spawn_information.gd
nalquas: Okay, seems good
```suggestion
var despawn_point: RoutePoint
```

DasCapschen:
And that's it. I've fixed the last few bugs in the Editor :)

... and doing so have completely re-written the Chunk loading. No more threads! 
Because we are using TSCN files now, I've made use of `ResourceLoader.load_interactive`, like the loading screen does.
Everything is on the main thread, but it's not blocking! And now it can actually be debugged! 🥳 

I've made sure everything works in the Editor and U2 Nuremburg. Feel free to give it a try.
(Note: in the Editor chunks can be a bit slower to load, but I have not had any problems in-game)

src/Editor/Modules/AddObjects.gd
```diff
@@ -1,5 +1,11 @@
 extends Control
 
+var editor
+
+
+func _ready():
+	editor = find_parent("Editor")
+
```
nalquas: I think this could be simplified using `onready`, no need to add the `_ready` function for just one variable.
```suggestion
onready var editor = find_parent("Editor")

```

nalquas:
Had a crash trying to save (Ctrl+S) in the scenario editor with no routes created:

`Invalid call. Nonexistent function 'duplicate' in base 'Nil'.` at `res://Editor/Scripts/ScenarioConfiguration.gd:44 - at function: save`

nalquas:
While in the scenario editor, a lot of the following error is being thrown:
```
E 0:03:37.103   decompose_polygon_in_convex: Convex decomposing failed!
  <C++ Source>  core/math/geometry.cpp:687 @ decompose_polygon_in_convex()
  <Stack Trace> Rail2DCollider.gd:56 @ _ready()
                scenario_map.gd:155 @ create_line2d_from_rail()
                scenario_map.gd:202 @ update_map()
                scenario_editor.gd:90 @ run_map_updater()
                scenario_editor.gd:27 @ _process()
```
I feel like that issue was already there before, though. It didn't affect anything visible to the user, anyway.

I don't think this has to be fixed in this PR, we could revisit this in a later PR.

nalquas:
Another quirk I noticed in an exported build of the game: Using the `Test track` button in the scenario editor fails with a message that you have to "create a scenario" to test, even though I've created one. Pressing `Test track` in the track editor, however, works flawlessly. Probably not related to this PR though (?), we might want to create an issue to revisit later.
Screenshot_20220828_140854

And another thing: I noticed the "Waiting Persons" setting in the RoutePoints menu isn't limited to the 0-100% range, is this a bug or a feature? Again, maybe not related to this PR.
Screenshot_20220828_135540_cropped

Edit: Looking over my comments again I think only the crash I mentioned above (https://github.com/Libre-TrainSim/Libre-TrainSim/pull/390#issuecomment-1229440291) has to be adressed in this PR, the other stuff is issue material for later.

DasCapschen:
> Had a crash trying to save (Ctrl+S) in the scenario editor with no routes created:

Nice one finding that 👍🏻 



> While in the scenario editor, a lot of the following error is being thrown:
> 
> ```
> E 0:03:37.103   decompose_polygon_in_convex: Convex decomposing failed!
>   <C++ Source>  core/math/geometry.cpp:687 @ decompose_polygon_in_convex()
>   <Stack Trace> Rail2DCollider.gd:56 @ _ready()
>                 scenario_map.gd:155 @ create_line2d_from_rail()
>                 scenario_map.gd:202 @ update_map()
>                 scenario_editor.gd:90 @ run_map_updater()
>                 scenario_editor.gd:27 @ _process()
> ```
> 
> I feel like that issue was already there before, though. It didn't affect anything visible to the user, anyway.
> 
> I don't think this has to be fixed in this PR, we could revisit this in a later PR.

This was already a problem. Something seems to be wrong with the code generating the colliders for the Rails. 
This also leads to the problem that some rails are not selectable.
I have replaced the single concave polygon by many rectangle shapes to approximate the line shape. Probably not the best solution, but is solves this problem for now.


> Another quirk I noticed in an exported build of the game: Using the `Test track` button in the scenario editor fails with a message that you have to "create a scenario" to test, even though I've created one. Pressing `Test track` in the track editor, however, works flawlessly. Probably not related to this PR though (?), we might want to create an issue to revisit later. 

This was already a bug. The scenario editor and the track editor used different paths trying to find the scenarios.
I have adjusted this behaviour, and it should work now.

> And another thing: I noticed the "Waiting Persons" setting in the RoutePoints menu isn't limited to the 0-100% range, is this a bug or a feature? Again, maybe not related to this PR.

I'm not sure about that one. I haven't worked on the PersonSystem. I don't even know why this number is in Percent to begin with. Probably best to leave it as is for now.

nalquas:
Thanks, fixes work for me.

However, something in b68391b3e513e74bf254c01308b241f21088e21c has resulted in the Editor not showing any objects except the track:

| Before (at 36945ef87ba891b124736e11dfdc60d10edfced3)      | After (at b68391b3e513e74bf254c01308b241f21088e21c) |
| ----------- | ----------- |
| Screenshot_20220828_162831      | Screenshot_20220828_162724       |

The objects show as normal when testing the track, though.

DasCapschen:
Indeed, I missed some path changes. Did a global search and should've fixed them all now. 
Works on my machine, feel free to test again :)

nalquas:
Yup, that fixed it.

*However*, the "Test track" button in the scenario editor is now differently broken. When loading the track from here, I get a grey screen and nothing happens.

https://user-images.githubusercontent.com/22664233/187080788-2a1e47d2-a7e8-4195-b0e9-924c1c765ea1.mp4

Using "Test track" in the track editor still works fine, btw.

nalquas:
Sorry, I seem to be to good at finding errors ;)
Creating a new world/track and trying to save results in the following:

```
E 0:00:07.551   load_interactive: Cannot open file 'user://editor/testtracktest/Worlds/testtracktest/testtracktest_config.tres'.
  <C++ Error>   Condition "err != OK" is true. Returned: Ref<ResourceInteractiveLoader>()
  <C++ Source>  scene/resources/resource_format_text.cpp:1183 @ load_interactive()
  <Stack Trace> World.gd:46 @ _ready()
                Editor.gd:727 @ load_world()
                Editor.gd:28 @ _ready()
                loading_screen.gd:127 @ _add_to_tree()
                loading_screen.gd:104 @ _clean_up_and_switch()
                loading_screen.gd:74 @ _process()
```

DasCapschen:
Don't be sorry. Every found error helps 😄 
I just found the issue with the missing `_config.tres` myself and fixed it. 

Found the bug with the scenario editor. If you were exporting, it wasn't correctly setting `Root.scenario_editor = false` ... 
That's why global variables are bad, mkay.

nalquas (dismissed):
Everything seems to be working now. Looks good from my side. :+1: 
Good job and thanks for all the fixes

HaSa1002:
The free driving scenario doesn't seem to end. It's winning condition is not triggered.

**Edit:** When departing too early

HaSa1002:
Overriding a red signal prompts the failure message but the scene is not switched on button click

HaSa1002:
Going over the end of the of the railroute after overriding the signal does terminate the play correctly.

HaSa1002:
The beginner tutorials don't show any text messages.
The signal tutorials seem to work though.

HaSa1002:
The saftety and signals tuorials do work but they don't have a speed set at the beginning

HaSa1002:
On load of an old map, I get the following errors:
```
E 0:25:44.945   surface_set_material: Index p_idx = 1 is out of bounds (surfaces.size() = 1).
  <C++ Source>  scene/resources/mesh.cpp:924 @ surface_set_material()
  <Stack Trace> Editor.gd:158 @ _port_to_new_chunk_system()
                Editor.gd:33 @ _ready()
                loading_screen.gd:127 @ _add_to_tree()
                loading_screen.gd:104 @ _clean_up_and_switch()
                loading_screen.gd:74 @ _process()
```
and
grafik

HaSa1002:
Clicking "New Rail" in the rail builder on a freshly created map, causes this wonderful bug:
grafik
It should be `origin` instead of `position`

Note: It also fails when adding a rail using the "+" Button

HaSa1002:
I just retested the tutorials in the main branch and they just as broken as they are here. Do you wanna try to fix them or do we move that to the future?

HaSa1002 (approved):
lgtm

HaSa1002:
Thanks ❤️

PR #277, opened 2021-12-05
Use Radians for Calculations

DasCapschen:
I changed all the calculations to use Radians instead of Degrees. This should make calculations a little faster because we don't need to make so many conversions. The UI still uses degrees and is the only place where conversions are used.

Also refactors some variable and function names to match the coding style.
Should not be considered done yet, because I think currently .tscn and .save files use Degrees, so I need to make sure to convert those when loading an old version, for backwards compatibility.

We should not merge this for 0.9, because it's too close to release

Depends on #276

PR #393, opened 2022-08-22
Fix: incorrect conversion to radians

DasCapschen:
With PR #277, I also converted `slope` and `tend` of the rail into radians, because I had incorrectly identified them as angles.  
This isn't the case, and I have now reverted the conversion, calculations with these values were never changed.  
Sorry for this error.

### DIFF
--- a/src/Data/Scripts/Rail.gd
+++ b/src/Data/Scripts/Rail.gd
@@ -23,8 +23,8 @@ export (Vector3) var end_pos: Vector3
 
 
 ## Steep
-export (float) var start_slope: float = 0  # Radians
-export (float) var end_slope: float = 0  # Radians
+export (float) var start_slope: float = 0  # % (meters / 100 meters)
+export (float) var end_slope: float = 0  # % (meters / 100 meters)
 
 export (float) var start_tend: float = 0
 export (float) var tend1_pos: float = -1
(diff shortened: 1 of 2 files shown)


DasCapschen:
Actually, tend is an angle, in degrees. But I do not fully understand the calculation, so I don't dare convert tend to radians, because it might break without adjusting the calculation. I just used deg2rad() for now.

That should've been all for the radians conversion PR post-mortem fixup 🙈

src/Data/Scripts/Rail.gd
```diff
@@ -199,7 +199,13 @@ func get_global_transform_at_distance(distance: float) -> Transform:
 # local to this rail
 func get_local_transform_at_distance(distance: float) -> Transform:
 	if parallel_rail_name == "":
-		return Transform(Basis().rotated(Vector3(1,0,0), get_tend_at_distance(distance)).rotated(Vector3(0,0,1), get_height_rot(distance)).rotated(Vector3(0,1,0), circle_get_rad(radius, distance)), get_local_pos_at_distance(distance) )
+		return Transform(\
+			Basis()\
+				.rotated(Vector3(1,0,0), get_tend_at_distance(distance))\
+				.rotated(Vector3(0,0,1), get_height_rot(distance))\
+				.rotated(Vector3(0,1,0), circle_get_rad(radius, distance)),\
+			 get_local_pos_at_distance(distance)\
```
HaSa1002:
```suggestion
		return Transform( \
			Basis() \
				.rotated(Vector3(1,0,0), get_tend_at_distance(distance)) \
				.rotated(Vector3(0,0,1), get_height_rot(distance)) \
				.rotated(Vector3(0,1,0), circle_get_rad(radius, distance)), \
			 get_local_pos_at_distance(distance) \
```
Please leave a space before the newline thingy.

HaSa1002 (dismissed):
lgtm. Feel free to merge when the style change is done.

HaSa1002 (dismissed):
lgtm. Please clean up the history

HaSa1002 (approved):
lgtm

PR #394, opened 2022-08-23
fix mod export missing textures and meshes

DasCapschen:
The modding tools were broken. 
The PCKPacker does not automatically pack imported resources like Textures and Meshes correctly.
The `.mesh` and `.stex` files were missing. (I believe this is why Jean once created https://github.com/Libre-TrainSim/Libre-TrainSim-Resource-Importer).
Anyways, the `.import` file is just a `ConfigFile` and can be manually read to get the paths of the imported files.

This should fix missing Track Screenshots and crashes due to missing chunks :)

### DIFF
--- a/src/addons/libre_trainsim_modding_tools/dock.gd
+++ b/src/addons/libre_trainsim_modding_tools/dock.gd
@@ -4,6 +4,8 @@ extends VBoxContainer
 var base: Control
 var dir_select_dialog: FileDialog
 
+var IMPORTED_RESOURCE_TYPES := ["StreamTexture", "Mesh"]
+
 func _on_new_mod_pressed() -> void:
 	var popup = preload("new_mod_popup.tscn").instance()
 	popup.base_control = base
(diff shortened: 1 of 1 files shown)


src/addons/libre_trainsim_modding_tools/dock.gd
```diff
@@ -45,24 +45,44 @@ func _on_export_dir_selected(dir: String) -> void:
...
+		Logger.err("Unable to copy content.tres to mod folder! (Reason: %s)" % err, self)
+
+
+func _get_imported_paths(file):
+	var cfg = ConfigFile.new()
+	if cfg.load(file) != OK:
+		Logger.err("cannot open .import file", self)
+	var type = cfg.get_value("remap", "type", "")
+	if type == "StreamTexture" or type == "Mesh":
+		return cfg.get_value("deps", "dest_files", [])
```
HaSa1002: Does this really only apply to these two? My gut tells me it's probably better to store the types in a const array (like BUGGING_PACK_TYPES) and check with in:
```suggestion
	if type in BUGGING_PACK_TYPES:
		return cfg.get_value("deps", "dest_files", [])
```

DasCapschen:
cleaned up history

HaSa1002 (approved):
lgtm

HaSa1002:
Cherrypicked for 0.9

PR #406, opened 2022-09-05
Fix Buildings Editing

HaSa1002:
- Fix building selection passing a node instead of a mesh
- Fix material name not shown when removing the default material


DasCapschen (approved):
lgtm

PR #428, opened 2022-10-03
Rework build menu

### LINKED ISSUE #421: Objects as TSCN, not Mesh
### Summary
When adding Objects to the Scene, you currently have to select the Mesh from a list and then apply materials 1-by-1 per surface.  
It would be nicer to select the finished object from a thumbnailed list.

### Proposal
Objects should be "prefabs". That is, pre-configured scenes (e.g. a MeshInstance) with already assigned materials.  
The user should not be allowed to select materials.  
Instead, if multiple variants using the same Object / Mesh are needed, multiple "prefabs" should be created with the same mesh but different materials assigned.  
This will also allow Thumbnails (Godot can already generate thumbnails for tscn files, perhaps we can somehow use this)

HaSa1002:
Pull Request is not yet ready. I aim to implement #421.

It has a bit of everything. A complete UX overhaul for the object addition coming with new button behaviour (A very hacky and painful process...), picked object moving, a palette like object placement, thumbnail generation (which is very hard to get right without user input), object search, object filtering, a resizer, and ofc a good amount of refactoring.

Missing:
 - [ ] **Colour palette** -> new PR and proposal.
 - [ ] Porting of the objects/Remake of the objects -> new PR

Regarding the porting of the objects: It may be better to just touch them once. Meaning, we port them and redo the meshes and materials in one shot. Otherwise we will port them to then redo the objects. This PR is going to be humongous.

For consideration: I didn't touch the rail logic stuff. I'd like to move them to the new system, too. However, we likely need some more features to get them working with nice UX.
[Video Showcase](https://user-images.githubusercontent.com/14185889/193704807-23999953-1e7f-4216-8b2f-a57e62e3210b.webm)

Bugs:
- [ ] Placing doesn't work sometimes because a deselction is fired <- Can't reproduce anymore.
- [x] Thumbnail generator is broken

Closes #421.


DasCapschen:
Oh moving Rail Logic stuff to the new system is a great idea!
We could incorporate snapping rail logics to rails while we're at it 😄

Do you also want to move Rails themselves to the new system?
Because if you remember, the system allows for different rail types (though I'm not sure that is exposed in the editor ; we only have 1 right now anyways), so that would be great for picking the right one!  

I understand this is going to be a huge amount of work.
Perhaps we would like to split this into multiple PRs? 
Especially porting objects and rail logic, that could be their own PRs.

Unfortunately, I will not be able to review this before 25th October :(

HaSa1002:
A little update on the colour tint stuff:
Unfortunately, we have to duplicate the materials in order to make that work because the material is shared across instances. (So far nothing new). However it results in one material per placed mesh. One optimisation strategy is storing the set colours per used mesh and only create a material per tinted colour. Ideally we would use per instance data but Godot doesn't support it in 3.x. The alternative being the usage of dedicated scenes isn't any better though.

HaSa1002:
> Oh moving Rail Logic stuff to the new system is a great idea!

I am honestly unsure about the UX in that case. We still wanna be able to access them quickly.

> We could incorporate snapping rail logics to rails while we're at it 😄

Yeah... Even though I have no idea how I'd solve that :D

> Do you also want to move Rails themselves to the new system?
Because if you remember, the system allows for different rail types (though I'm not sure that is exposed in the editor ; we only have 1 right now anyways), so that would be great for picking the right one!

Yes, but probably as their own menu? I am not sure, yet.

> I understand this is going to be a huge amount of work.
Perhaps we would like to split this into multiple PRs?
Especially porting objects and rail logic, that could be their own PRs.

Yeah. The objects actually pose some more issues. We will break all the old maps again.
The porting itself is not so much the problem but fixing the objects is. I tried fixing the office_1 yesterday and I hit some issues:
First, it took me about two or three hours to clean up the mesh and starting multiple texturing attempts. Second, I am not happy at all how the mesh looks (it resembles a mental image of an office building but real life looks waaaaaay different). Third, the texturing didn't went smooth at all. Sure, a good amount of time was wasted finding a workflow and fooling around with the tools but I super unsure how to go about it. Fourth, there are about a hundred meshes and fifth, we break old maps if we decide to export the meshes not as obj or move them. I would like to take the opportunity to clean up the objects folder and introduce subfolder to aid search. However, I don't wanna fix every single reference in Hainfurt, tutorials, and Nuremberg by hand.

> Unfortunately, I will not be able to review this before 25th October :(

Take your time :) It's a hobby :>

HaSa1002:
Another update on the colours:
I want to add a colour palette to enable colour randomisation that is actually useful and to speed up the tinting workflow. We may wanna provide some default palettes.
Here's a new mockup/wireframe:
grafik

Description of the colour palette features:
1. Support for up to 24(?) colours -> verify good number
2. Multiple colours can be selected
3. Next colour is randomisable (green button)
4. Menu Button lists all “installed” colour palettes
5. Hovering over a palette shows the palette
6. Creating a new palette opens the modal window
7. Usage of default Godot colour picker
8. Showcase of the added colours
9. Each colour is put into the text field in hex
10. Edits of the text field are immediately parsed

DasCapschen:
Tbh, I'm not quite sold on the color tinting feature.  
For me, improving the Editor Workflow is already a huge step even without this "gimmick", since tinting is not nearly the same as having a different material that makes the object look fundamentally different.  
So perhaps, if this is making so many problems, we should move it to it's own PR and deal with it in the future?  
Let's try to get the UX right before adding more features?

About Rails and Rail Logic: these could be their own Tab in the build menu?  
Like you have "Buildings", "Vegetation", "Rail Objects", "Rails", "Favourites", etc.
If it becomes too cluttered, we could introduce sub-tabs.
Alternatively, I think the search should provide quick enough access to anything.

Don't worry about rail snapping, we'll figure it out, can (and should) go into it's own PR :)

Since you said you're not happy with how the office looks for example: I'm not sure completely redesigning and replacing old objects is the way to go. We could just make new objects that look more realistic, and do the minimal amount of work needed to make old objects work with the new system (or at least not break old maps)
Optimizing their meshes (and materials) might still be a good idea. But I won't leave you alone to all meshes (unless you wanna do it alone lol)

Cleaning up the object folder sound like a good idea.
About fixing references in maps: there is always a text editor and search & replace. We're using **T**SCN for a reason :)
Already used that trick plenty to fix up U2 and Hainfurt for TSCN chunks!

HaSa1002:
> Tbh, I'm not quite sold on the color tinting feature.
For me, improving the Editor Workflow is already a huge step even without this "gimmick", since tinting is not nearly the same as having a different material that makes the object look fundamentally different.

You're right. The tinting is indeed a rather small detail. However, it is the way the various wall materials work. The goal is to add some easy variety.

> So perhaps, if this is making so many problems, we should move it to it's own PR and deal with it in the future?

I made a working version using the outlined optimisation proposal. So *something* is there. It's just not as useful as I'd like it to be. Nor as usable. The proposed solution enables colour randomisation that makes sense. All other proposed changes in that mockup aim to improve the usability.

> Let's try to get the UX right before adding more features?
I agree.

> About Rails and Rail Logic: these could be their own Tab in the build menu?
Like you have "Buildings", "Vegetation", "Rail Objects", "Rails", "Favourites", etc.
If it becomes too cluttered, we could introduce sub-tabs.
Alternatively, I think the search should provide quick enough access to anything.

I am thorn. On one hand, I like having everything in one menu. However, they may require new filters/tabs to be added (though I'd begin with adding them into infrastructure). The downside of the filters *in* the objects menu is the reduced hierachy. Meaning, we essentially move the "top level filters" (that are buttons right now) next to all other objects. I think we loose efficency by doing so.
My initial idea was reusing the objects menu for rails and logic but only with relevant content.

> Don't worry about rail snapping, we'll figure it out, can (and should) go into it's own PR :)

Yeah. Never intended doing it here xD

> Since you said you're not happy with how the office looks for example: I'm not sure completely redesigning and replacing old objects is the way to go. We could just make new objects that look more realistic, and do the minimal amount of work needed to make old objects work with the new system (or at least not break old maps)

Yeah so here's the catch. What do we want? This PR will not break any map because the meshes are referenced in the worlds. However, you won't be able to build anything if we don't port some objects. Creating new seems like the way to go (these can be put in a separate repo and pack file probably).

The breaking change (going forward) is going to be the attached rail buildings. Going away from objs is going to break *that*. Not of importance for this particular change, though.

> Optimizing their meshes (and materials) might still be a good idea. But I won't leave you alone to all meshes (unless you wanna do it alone lol)

Hehe.

> Cleaning up the object folder sound like a good idea.

See note above.

> About fixing references in maps: there is always a text editor and search & replace. We're using TSCN for a reason :)
Already used that trick plenty to fix up U2 and Hainfurt for TSCN chunks!

That's high level wizardry.

HaSa1002:
So I'd say, let's have a review run and try it. We need to get it out of my hands for two reasons. First, my time is unpredictable right now and it needs usage to find better solutions. I designed the feature from very personal perspective and that may work as a starting ground but is in no way sustainable and scalable.

Please try, please find bugs. I may actually convert a lot of those objects to scenes, just to have them available for testing purposes.

src/Data/Scripts/Resources/editor_info.gd
```diff
@@ -0,0 +1,20 @@
+class_name EditorInfo
+extends Resource
+
+
+const RECENT_QUEUE_SIZE := 20
+
+
+export var favourites := {} # scene: null
```
DasCapschen, line 8: Why isn't this just an Array? 😅

src/Data/Scripts/Singletons/Root.gd
```diff
@@ -150,10 +150,6 @@ func get_subfolders_of(directory_path: String) -> Array:
 	return folder_names
 
 
-func set_fullscreen(value: bool) -> void:
-	OS.window_fullscreen = value
-
-
```
DasCapschen, line 156: Why is this deleted?

src/Editor/Docks/RailAttachments/RailAttachments.gd
```diff
@@ -254,7 +254,9 @@ func _on_PickObject_pressed() -> void:
 
 
 func _on_Content_Selector_resource_selected(complete_path: String) -> void:
-	$Tab/TrackObjects/Settings/Tab/Object/BuildingSettings._on_Content_Selector_resource_selected(complete_path)
+	#$Tab/TrackObjects/Settings/Tab/Object/BuildingSettings._on_Content_Selector_resource_selected(complete_path)
+	print("Do not forget to remove the content selector callback!")
```
DasCapschen: So, why is it still here? 😜

src/Editor/Scripts/Editor.gd
```diff
@@ -712,7 +712,16 @@ func load_world() -> bool:
 		send_message("World data could not be loaded! Is your .tscn file corrupt?\nIs every resource available?")
 		return false
 
-	var world: Node = world_resource.instance()
+	world = world_resource.instance() as LTSWorld
+	if !world:
+		send_message("Failed to load world. World is not an LTSWorld.")
+		return false
+
+	editor_info = load(current_track_path.plus_file("editor_info.tres"))
```
DasCapschen, line 728: Are you sure that Editor Info should be saved per-track? Not rather per-User? 🤔

src/Editor/ui/icons/star.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="34" height="31" viewBox="0 0 34 31" fill="none" xmlns="http://www.w3.org/2000/svg">
```
DasCapschen, line 1: Why is this the only icon with an SVG? 🤔

src/Editor/ui/objects_menu/objects_menu.gd
```diff
@@ -0,0 +1,238 @@
...
+var moving_camera := 0
+var select_random := false
+
+var filter_building := false
+var filter_vegetation := false
+var filter_infrastructure := false
+var filter_decorative := false
+var filter_favourites := false
+var filter_common := false
+var filter_recent := false
```
DasCapschen, line 26: Can you filter multiple things at once?
Because if not, it's a state machine and should use an Enum.

src/Editor/ui/objects_menu/objects_menu.gd
```diff
@@ -0,0 +1,238 @@
...
+	assert(editor_info)
+
+	find_groups()
+	populate_content()
+	generate_thumbnails()
+
+
+func _process(delta: float) -> void:
+	if current_object == null || !current_object in selected_objects:
+		select_object(0)
```
DasCapschen: Why?

src/Editor/ui/size_changer.gd
```diff
@@ -0,0 +1,53 @@
...
+onready var target := get_node(target_path) as Control
+
+
+func _ready() -> void:
+	assert(target)
+	match drag_side:
+			Side.Right, Side.Left:
+				mouse_default_cursor_shape = Control.CURSOR_HSIZE
+			Side.Top, Side.Bottom:
+				mouse_default_cursor_shape = Control.CURSOR_VSIZE
```
DasCapschen: Too much indented (2 indents after match)

src/Editor/ui/size_changer.gd
```diff
@@ -0,0 +1,53 @@
...
+extends Separator
+
+
+enum Side {
+	Top = MARGIN_TOP,
+	Left = MARGIN_LEFT,
+	Bottom = MARGIN_BOTTOM,
+	Right = MARGIN_RIGHT
+}
+
```
DasCapschen: Should enums always be CAPITAL?

src/addons/jean28518.jTools/jSettings/JSettings.gd
```diff
@@ -31,7 +31,7 @@ func first_run_check():
 
 func reset_settings_to_default():
 	Logger.log("Settings reset to default.")
-	set_fullscreen(true)
+	#set_fullscreen(true)
```
DasCapschen: Why is this gone?

DasCapschen (changes requested):
Code looks alright, have not tested it in-action yet.
Just some questions, see comments.

src/Data/Scripts/Resources/editor_info.gd
HaSa1002: It's used to perform check logic (scene in favourites). If I'd use an array, we'd have O(n) instead of O(1) for the look up. I essentially use the dict as hashset. 😅

src/Editor/Docks/RailAttachments/RailAttachments.gd
HaSa1002: ehhh. There was ... *a reason*

src/Editor/Scripts/Editor.gd
HaSa1002: yeah. Right now it is just an reference to keep scene specific data anyway. If we expand more data into it, we may wanna push it to be user specific. I feel like the maps may be vastly different and having different favourites makes sense

HaSa1002: It used to store the colour combinations. But I removed that

src/Editor/ui/icons/star.svg
HaSa1002: They all are vector images, but Godots importer sucks. I think, I am gonna push the other icons as svg as well. But we do need the PNGs for now.

src/Editor/ui/objects_menu/objects_menu.gd
HaSa1002: yes, you can

src/Editor/ui/objects_menu/objects_menu.gd
HaSa1002: The building sticks to your cursor. If you are in the menu and deselect the building, it needs to update the building. I simply select the first one in the list.

src/Editor/ui/size_changer.gd
HaSa1002: https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#constants-and-enums
...
oh. I'm gonna fix that

src/Data/Scripts/Singletons/Root.gd
HaSa1002: Because of the very first commit. With that in place, it doesn't respect the project settings

src/addons/jean28518.jTools/jSettings/JSettings.gd
HaSa1002: Because of the very first commit. With that in place, it doesn't respect the project settings

DasCapschen: but won't that affect the exported version as well?

src/Editor/ui/icons/star.svg
DasCapschen: I thought Godot can use SVGs, why do we need the PNGs? 🤔 
But it's okay. I was just wondering why this is the only one with an svg.

src/Editor/ui/objects_menu/objects_menu.gd
DasCapschen: But then something is always selected. Shouldn't it be possible to have nothing selected?

src/Editor/Scripts/Editor.gd
DasCapschen: Hmm, I guess it's okay for now. But I bet users will wonder where their favourites went when making a second map.

src/Editor/ui/icons/star.svg
HaSa1002: yeah, but they are rasterised using a library that works sometimes

src/Editor/ui/objects_menu/objects_menu.gd
HaSa1002: You can deselect everything by pressing RMB

DasCapschen: I guess I don't quite understand the code then.
Because on RMB, you set `current_object = null`, but here, if it is `null` you select something.

Oh, I guess `entries[entry].deselect_all()` somehow affects `selected_objects`?
If so, that is not obvious at all! Why should calling a function on some object affect a dictionary here? That sounds more like a nasty side-effect than nice code structure.
At the very least add a comment explaining this.

HaSa1002: The decision depends on a time threshold because I need to determine wether or not you wanted to deselect the selection or pressed RMB to move the camera.

DasCapschen: The longer I look at this code, and the more you explain it, the worse it gets 🙈 
Maybe let's have a call on Discord tomorrow or something...

HaSa1002: See
https://github.com/Libre-TrainSim/Libre-TrainSim/pull/428/files#diff-561ae66c03d0c7eb078ec707744ce7e0b393272244d719f0154c98d38557b48fR141

and

https://github.com/Libre-TrainSim/Libre-TrainSim/pull/428/files#diff-561ae66c03d0c7eb078ec707744ce7e0b393272244d719f0154c98d38557b48fR177-R181

HaSa1002: yeah sure

src/Data/Scripts/Singletons/Root.gd
HaSa1002: it was actually unused

DasCapschen: well in that case 😄

src/addons/jean28518.jTools/jSettings/JSettings.gd
HaSa1002: All fixed in the latest commit

DasCapschen: Looks good! :+1:

src/Editor/ui/size_changer.gd
HaSa1002: fixed

src/Editor/ui/size_changer.gd
HaSa1002: fixed

src/Editor/Docks/RailAttachments/RailAttachments.gd
HaSa1002: fixed

HaSa1002:
Materials are changable again

Edit: But not in prefabs.

HaSa1002:
I fixed the material selection. It's working properly now.

HaSa1002:
Maybe, we could save prefab variants using the building settings? Doesn't need to be this PR. Maybe we should wait for feedback.
Just an idea, I just had.

¯\\_(ツ)_/¯

DasCapschen:
> Maybe, we could save prefab variants using the building settings? 

You mean like people place an object and select their wanted materials and can then save this Object-Material-Combo to their user settings, so they can use it again later?
That actually sounds like a good idea.

src/Editor/ui/objects_menu/objects_menu.gd
HaSa1002: I fixed the camera movement detection logic

HaSa1002:
Yeah exactly.

HaSa1002:
Please take another look. All shoud be working now

DasCapschen (approved):
lgtm

PR #432, opened 2022-11-26
Fix various bugs

DasCapschen:
This should fix issues with rail routes that @nalquas has found in U2 https://github.com/Libre-TrainSim/u2-nuremberg/pull/4#issuecomment-1242973492

Also fixes some errors in the Grey Train, there were some wrong Node paths.

Also fixes some small AI bugs. Despawning is still broken though :(

Also fixes KS Signals displaying speeds when red (they are not allowed to do that IRL)


HaSa1002 (approved):
lgtm

PR #434, opened 2022-11-26
Fix descriptions not being displayed

HaSa1002:
We have like two descriptions. So I load the one that is set...


DasCapschen (approved):
lgtm 👍🏻 
Thank you ❤️

PR #435, opened 2022-11-27
Fix lost auto pilot

HaSa1002:
I tried to find causes for the bugs we encountered in Nuremberg. I did fix one major issue, however, I think most issues in nuremberg are scheduling issues in combination with the block system not working for trains moving in different directions. I guess we have to swallow that pill for now and ensure there are some safety margins.
**Note:** The last npc train in nuremberg doesn't leave Röthenbach and I have no idea why.

I also added a debug camera, to be able to follow trains (F12 to open the menu, Shift F12 to free the cursor).
As I am unpatient, I added a time scale debug option. It works and eases debugging so much.

NPC1 has a negative speed, thus a negative distance on rail and thus causes an "immediate" despawn. The max is just a duct tape but hides the issue.

### DIFF
--- a/src/Data/Scripts/HUD.gd
+++ b/src/Data/Scripts/HUD.gd
@@ -95,7 +95,7 @@ func update_nextTable() -> void:
 		$IngameInformation/Next/GridContainer/DistanceToSpeedLimit.text = "-"
 
 	## Update Next Station
-	if player.station_table.size() == 0 or (player.current_station_table_entry.stop_type == StopType.END and player.is_in_station):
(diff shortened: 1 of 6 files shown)


DasCapschen (approved):
Looks good for now.
I'll investiage these issues more next week.
Thanks a lot for the debugging tools 👍🏻

Jean28518:
I highly recommend to redo the scenarios with the new scenario editor. (the old scenarios shouldn't be compatible with the new scenario editor and the new scenario handler) (It is a lot easier with the new scenario editor to create a new scenario)
I can do this if you don't want to get into this.

nalquas (approved):
Lgtm!

Jean28518 (approved):

PR #436, opened 2022-11-27
Fix end - end rail connector

Jean28518:
End - End Rail Connections didn't work anymore.

### DIFF
--- a/src/Editor/Scripts/Editor.gd
+++ b/src/Editor/Scripts/Editor.gd
@@ -514,7 +514,7 @@ func handle_drag_mode() -> void:
 			snap_rot = snapped_object.start_rot
 			snap_pos = snapped_object.start_pos
 		else:
-			snap_rot = snapped_object.end_rot - deg2rad(180.0)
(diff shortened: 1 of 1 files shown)


src/Editor/Scripts/Editor.gd
```diff
@@ -514,7 +514,7 @@ func handle_drag_mode() -> void:
 			snap_rot = snapped_object.start_rot
 			snap_pos = snapped_object.start_pos
 		else:
-			snap_rot = snapped_object.end_rot
+			snap_rot = snapped_object.end_rot - deg2rad(180.0)
```
DasCapschen: use `PI` instead of `deg2rad(180)`

Otherwise thanks for the fix :+1:

DasCapschen (changes requested):

src/Editor/Scripts/Editor.gd
Jean28518: Done. thanks for reviewing!

DasCapschen (approved):
lgtm

PR #440, opened 2022-11-27
Fix _temp chunk cleaning

### LINKED ISSUE #437: Editor: Changes sometimes are not saved
when `_temp.tscn` exists of a chunk, new changes won't be saved...

Jean28518:
for files in user:// space you need to call _dir.open() first :)

Fixes #437

### DIFF
--- a/src/Data/Scripts/chunk_manager.gd
+++ b/src/Data/Scripts/chunk_manager.gd
@@ -310,6 +310,7 @@ func cleanup():
 	var files_to_remove := []
 
 	var chunk_path = editor.current_track_path.plus_file("chunks")
+	_dir.open(chunk_path)
 	_dir.change_dir(chunk_path)
 	_dir.list_dir_begin(true, true)
 	while(true):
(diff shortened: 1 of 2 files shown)


src/Data/Scripts/chunk_manager.gd
```diff
@@ -310,6 +310,7 @@ func cleanup():
 	var files_to_remove := []
 
 	var chunk_path = editor.current_track_path.plus_file("chunks")
+	_dir.open(chunk_path)
```
DasCapschen, line 313: It seems weird to me that this is necessary, because `_dir.open()` is called in `_ready`, so only the `change_dir()` should be necessary... but I guess it doesn't hurt

DasCapschen (approved):

src/Data/Scripts/chunk_manager.gd
Jean28518: I agree.  I guess it depends on the dir object.
Without this line dir read this dir read res://.

PR #443, opened 2022-12-04
Update "distance" description in rail logic editor

### LINKED ISSUE #222: Distance is actually an offset
The parameter "Distance" shown in the image below is not a distance, but rather an offset.
grafik

Jean28518:


Jean28518:
Nothing changed at the layout. Godot did this automatically.

HaSa1002 (approved):
lgtm

PR #451, opened 2022-12-27
Fix track objects not updating their position on world origin shift

### LINKED ISSUE #441: Big Chunk System removes track objects
### Summary
Every time a big chunk change happens, the current loaded landscape gets moved to nowhere...
In the editor everything works fine. I only ca reproduce this ingame...

https://user-images.githubusercontent.com/39700889/205491444-a3a286b8-2e96-433a-b2a4-aa3d96572ec8.mp4

HaSa1002:
I guess the rails updated the TOs at some point which may have been removed by one of us because it is unbenefical coupling of classes.

Fixes #441

### DIFF
--- a/src/Data/Scripts/TrackObjects.gd
+++ b/src/Data/Scripts/TrackObjects.gd
@@ -113,6 +113,7 @@ func _exit_tree() -> void:
 
 func _ready() -> void:
 	make_mesh_unique()
+	Root.connect("world_origin_shifted", self, "update")
 
 
 func make_mesh_unique():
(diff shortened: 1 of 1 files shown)


DasCapschen:
No this should not be necessary. TrackObjects are inside the chunks, and iirc the whole chunk is shifted / the TrackObject parent node in the chunk is shifted.
It should not be necessary to shift every TrackObject individually (which is also very imperformant iirc)

HaSa1002:
Sorry misclick

HaSa1002:
> No this should not be necessary. TrackObjects are inside the chunks, and iirc the whole chunk is shifted / the TrackObject parent node in the chunk is shifted.

Nope we don't move the chunk. We move the buildings and the grass. The TO parent is not moved because you encountered some issue (?) I belive at some point the rails called the update of the rail attachements but that is no more.

https://github.com/Libre-TrainSim/Libre-TrainSim/blob/master/src/Data/Scripts/chunk.gd#L69-L81

> It should not be necessary to shift every TrackObject individually (which is also very imperformant iirc)

Well, I agree but it never worked that way.

DasCapschen:
Oh, you are right, I remember now... Track Objects really need a refactor at some point I think 🙈

DasCapschen (approved):
lgtm

PR #452, opened 2022-12-27
Fix grass and chunk not agreeing on the correct coordinate system to use

### LINKED ISSUE #442: Grass Planes are loading to late sometimes
### Summary
Grass Planes are loading to late sometimes

### How to reproduce
Editor or ingame: Fly or drive for long distances

### Solution:
Lets create a grass plane per chunk. (If this is already happing this bug happens because of: #441 

(Ignore the missing landscape in the video)

https://user-images.githubusercontent.com/39700889/205491664-89a2076e-dd6d-4147-86e6-58c3139ecbc6.mp4

HaSa1002:
... Which leads to missing grass planes.

We should add some math to the chunk itself and use that for all calculations.

Fixes #442

### DIFF
--- a/src/Data/Scripts/chunk_manager.gd
+++ b/src/Data/Scripts/chunk_manager.gd
@@ -88,6 +88,18 @@ func _ready():
 		world.add_child(chunks_node)
 		chunks_node.owner = world
 
+	yield(get_tree(), "idle_frame")
+	# get position of active camera
+	var position_provider = get_viewport().get_camera()
+	if position_provider == null:
+		Logger.err("Failed to perform initial move", self)
+		return
+
+	# handle world origin
+	var position = position_provider.global_transform.origin
+	var chunk_position = position_to_chunk(position)
+	_shift_world_origin_to(-chunk_position * chunk_size)
+
 
 func _order_rails_by_chunk():
 	for rail in world.get_node("Rails").get_children():
--- a/src/Data/Scripts/chunk_manager.gd
+++ b/src/Data/Scripts/chunk_manager.gd
@@ -40,18 +40,22 @@ static func string_to_chunk(chunk: String) -> Vector3:
 	return Vector3(x, 0, z)
 
 
-func get_3x3_chunks(around: Vector3):
-	return [
-		chunk_to_string(Vector3(around.x - 1, 0, around.z)),
-		chunk_to_string(Vector3(around.x - 1, 0, around.z - 1)),
-		chunk_to_string(Vector3(around.x - 1, 0, around.z + 1)),
-		chunk_to_string(Vector3(around.x + 1, 0, around.z)),
-		chunk_to_string(Vector3(around.x + 1, 0, around.z - 1)),
-		chunk_to_string(Vector3(around.x + 1, 0, around.z + 1)),
+func get_chunks(around: Vector3, distance: int):
+	var chunks := [
 		chunk_to_string(Vector3(around.x, 0, around.z)),
-		chunk_to_string(Vector3(around.x, 0, around.z - 1)),
-		chunk_to_string(Vector3(around.x, 0, around.z + 1)),
 	]
+	for i in range(1, distance + 1):
+		chunks.append_array([
+			chunk_to_string(Vector3(around.x - i, 0, around.z)),
+			chunk_to_string(Vector3(around.x - i, 0, around.z - i)),
+			chunk_to_string(Vector3(around.x - i, 0, around.z + i)),
+			chunk_to_string(Vector3(around.x + i, 0, around.z)),
+			chunk_to_string(Vector3(around.x + i, 0, around.z - i)),
+			chunk_to_string(Vector3(around.x + i, 0, around.z + i)),
+			chunk_to_string(Vector3(around.x, 0, around.z - i)),
+			chunk_to_string(Vector3(around.x, 0, around.z + i)),
+		])
+	return chunks
 
 
 func is_position_in_loaded_chunk(position: Vector3):
@@ -111,8 +115,8 @@ func _process(_delta: float):
 
 	# handle chunks
 	if chunk_position != active_chunk:
+		loader.load_chunks(get_chunks(chunk_position, ProjectSettings["game/gameplay/chunk_load_distance"]))
 		active_chunk = chunk_position
-		loader.load_chunks(get_3x3_chunks(active_chunk))
 		_unload_old_chunks()
 
 	if ProjectSettings["game/debug/display_chunk"]:
@@ -250,7 +254,7 @@ func resume_chunking():
 	set_process(true)
 	loader.set_process(true)
 	if active_chunk:
-		loader.load_chunks(get_3x3_chunks(active_chunk))
+		loader.load_chunks(get_chunks(active_chunk, ProjectSettings["game/gameplay/chunk_load_distance"]))
 
 
 func _save_chunk(chunk_name: String, saving: bool = false):
--- a/src/Data/Scripts/chunk_manager.gd
+++ b/src/Data/Scripts/chunk_manager.gd
@@ -18,7 +18,7 @@ var _dir: Directory = null
 
 func position_to_chunk(position: Vector3) -> Vector3:
 	position = position - world_origin
-	return Vector3(int(position.x / chunk_size), 0, int(position.z / chunk_size))
+	return Vector3(round(position.x / chunk_size), 0, round(position.z / chunk_size))
 
 
 func chunk_to_position(chunk: Vector3) -> Vector3:
@@ -65,6 +65,7 @@ func _ready():
 	if Root.Editor:
 		editor = find_parent("Editor")
 		assert(editor != null)
+		_test_position_calc()
 
 	_dir = Directory.new()
 	if _dir.open("res://") != OK:
@@ -189,6 +190,11 @@ func _force_load_chunk_immediately(chunk_pos: Vector3):
 	return loader._force_load_chunk_immediately(chunk_name)
 
 
+func _force_load_chunk_name_immediately(chunk_name: String):
+	assert(Root.Editor)
+	return loader._force_load_chunk_immediately(chunk_name)
+
+
 # TODO: Godot Export "TrackObject is not a valid type"
 func add_track_object(track_object: TrackObject):
 	var rail_name = track_object.attached_rail
@@ -243,7 +249,8 @@ func resume_chunking():
 
 	set_process(true)
 	loader.set_process(true)
-	loader.load_chunks(get_3x3_chunks(active_chunk))
+	if active_chunk:
+		loader.load_chunks(get_3x3_chunks(active_chunk))
 
 
 func _save_chunk(chunk_name: String, saving: bool = false):
@@ -328,3 +335,22 @@ func cleanup():
 
 	_dir.change_dir("res://")
 
+
+func _test_position_calc() -> void:
+	var cases := PoolVector3Array([
+		Vector3(0, 0, 0),
+		Vector3(500, 0, 0),
+		Vector3(-499, 0, 152),
+		Vector3(-501, 0, 0),
+		Vector3(1000, 0, -1000),
+		Vector3(499, 0, 1000),
+		Vector3(600, 0, 300),
+		Vector3(-700, 0, -700),
+	])
+
+	Logger.warn("Running position test", self)
+
+	for case in cases:
+		printt(case, position_to_chunk(case))
+
+	print(Vector3() == position_to_chunk(Vector3(-499, 0, 152)))
--- a/src/Editor/Scripts/Editor.gd
+++ b/src/Editor/Scripts/Editor.gd
@@ -37,6 +37,7 @@ func _ready() -> void:
 	Root.connect("world_origin_shifted", self, "_on_world_origin_shifted")
 
 	_port_to_new_chunk_system()
+	_port_v1_to_v2_chunks()
 	_port_to_new_scenario_system()
 	_port_very_old_scenarios()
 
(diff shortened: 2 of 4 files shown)


src/Data/Scripts/chunk.gd
```diff
@@ -31,6 +31,7 @@ func _ready() -> void:
 			add_child(default_grass)
 			default_grass.owner = self
 		$DefaultGrass.translation = chunk_position * 1000  # 1000 = ChunkManager.chunk_size
+		$DefaultGrass.translation += Vector3(500, 0, 500) # because the default grass thinks the coordinates of the chunk is in the centre point, but it is actually not.
```
DasCapschen: 1. you could've just changed the DefaultGrass in the scene? 🤔 
2. The chunk coordinates are at the chunk centre... chuck (0,0,0) should have grass in all 4 quadrants 🤔

HaSa1002: > you could've just changed the DefaultGrass in the scene? 🤔

Yes, however, Hainfurt and the tutorials all'd need to be resaved as the node gets saved inside the chunk.

> The chunk coordinates are at the chunk centre... chuck (0,0,0) should have grass in all 4 quadrants 🤔

No they are not. See this function: Perhpas the chunk position deciphering should have been inside the chunk. Dunno if you can actually access the class to get access to perhaps static methods.
https://github.com/Libre-TrainSim/Libre-TrainSim/blob/master/src/Data/Scripts/chunk_manager.gd#L19-L21

DasCapschen: `Vector3(int(position.x / chunk_size), 0, int(position.z / chunk_size))`

if `position.x` is between `-chunksize .. chunksize`, the result is 0 ; same for `position.z`

ex.: chunk size is 1000
if position = (0,0,0), chunk is (0,0,0)
if position = (600, 0, 300), chunk is (0,0,0)
if position = (-700, 0, -700), chunk is (0,0,0)

-> chunk position is the centre of the chunk

HaSa1002: So chunk_size is not 1000 but 2000 metres which would explain why there are missing grass planes.

HaSa1002: Maybe I am just super stupid right now. Idk :/

DasCapschen: You're onto something, the chunk position is *not* the centre... but it's very weird... I made a sketch
grafik

DasCapschen: nvm, that was wrong, lemme try again xD

HaSa1002: xD
The chunks overlap and that's why the issues was never apparent until you get to a corner case where it doesn't load the new chunk.

DasCapschen: So this is how it *should* work, this is the idea behind it:
- has no overlap
- extend of chunk in each direction is `chunk_size / 2`
- I feel like we've forgotten the `/2` part somewhere maybe? 🤔 
grafik

HaSa1002: but... Idk what we do with that knowledge (if it is correct)

HaSa1002: > So this is how it _should_ work, this is the idea behind it:
> 
>     * has no overlap

This is the part I am most worried about. We likely break our maps by fixing it (and we do need to fix it). I think, it will actually work most of the time because Jean heavily used rail attachments whoose position is derived from their rail. However, the maps will not load properly, so what do we do?

>     * extend of chunk in each direction is `chunk_size / 2`
>
>     * I feel like we've forgotten the `/2` part somewhere maybe? 🤔

Yeah I agree.

DasCapschen: well, if we are saying that the `chunk_position` is the centre of the chunk, then it should be calculated as `chunk_position = Vector( int(position.x/chunk_size*2), 0, int(position.z/chunk_size*2) )`
🤔 

and because the grass position is `chunk_position * chunk_size`, but `chunk_position` is wrong... the grass position is wrong!
Well, for chunk (1,0,0) the grass position becomes (1000, 0, 0)
But because we forget to that the chunk position is the centre, and didn't multiply by 2, the chunk starts at 1000, not 500 ; and instead of overlap, we actually have GAPS between the chunks? 🤯 

or maybe I've got something wrong here...

HaSa1002: If it would work, we wouldn't have overlaps because the grass is position around (0|0|0). But because of the bug, we have gaps of 500 metres when the chunk gets generated on the fly in one direction because it is empty.

What I don't get is why we never used Vector3.snapped tbh. It would do the propper rounding.
Here's a geogebra file for visual representation:
[LTS Chunks.ggb.txt](https://github.com/Libre-TrainSim/Libre-TrainSim/files/10329138/LTS.Chunks.ggb.txt)
(you have to remove the `.txt` part... GH didn't allow me uploading the file without it...)

DasCapschen: Yup, exactly, that's the issue, so that's what we have to fix. Do you want to do it in this PR? And feel free to use `Vector3.snapped`. I didn't use it because I didn't know it existed... 🙈

HaSa1002: I am going to leave this PR open for now. Maybe I am doing it here, maybe in a new PR. I'll have to think about a migration strategy for a bit. I really hope, we can simply push the fix and be lucky that nothing will break... idk.

DasCapschen: You could try to fix it and check if any of the maps break? I'm not sure if they would... 🤔

HaSa1002:
I think there's still something wrong with the chunks https://youtu.be/SQfjWXMiTqE

DasCapschen:
> I think there's still something wrong with the chunks https://youtu.be/SQfjWXMiTqE

Could be a Loader Problem? Maybe the interactive loader doesn't get enough time to stream in the assets?  
Are we making sure we load 2 chunks ahead, instead of just 1? That should also give enough buffer to ensure this doesn't happen 🤔 

Or is it actually the Chunk Manager not spawning in the right chunk / at the right position?

HaSa1002:
[No, we just load one chunk ahead afaik.](https://github.com/Libre-TrainSim/Libre-TrainSim/blob/master/src/Data/Scripts/chunk_manager.gd#L112-L115) That'll fix this issue probably. The streaming itself shouldn't be an issue tbf.

HaSa1002:
Should be working now

src/Data/Scripts/chunk_manager.gd
```diff
@@ -18,7 +18,7 @@ var _dir: Directory = null
 
 func position_to_chunk(position: Vector3) -> Vector3:
 	position = position - world_origin
-	return Vector3(int(position.x / chunk_size), 0, int(position.z / chunk_size))
+	return Vector3(round(position.x / chunk_size), 0, round(position.z / chunk_size))
```
DasCapschen, line 21: Didn't you want to use Vector.snapped?

src/Data/Scripts/chunk_manager.gd
```diff
@@ -83,6 +88,18 @@ func _ready():
...
+	# get position of active camera
+	var position_provider = get_viewport().get_camera()
+	if position_provider == null:
+		Logger.err("Failed to perform initial move", self)
+		return
+
+	# handle world origin
+	var position = position_provider.global_transform.origin
+	var chunk_position = position_to_chunk(position)
+	_shift_world_origin_to(-chunk_position * chunk_size)
```
DasCapschen, line 101: That's a good idea! 😄

src/Data/Scripts/chunk_manager.gd
```diff
@@ -110,8 +127,8 @@ func _process(_delta: float):
 
 	# handle chunks
 	if chunk_position != active_chunk:
+		loader.load_chunks(get_chunks(chunk_position, ProjectSettings["game/gameplay/chunk_unload_distance"]))
```
DasCapschen: Should we really re-use `chunk_unload_distance` for loading?
Perhaps just introduce a second variable `chunk_loading_distance`?

src/Editor/Scripts/Editor.gd
```diff
@@ -36,6 +36,7 @@ func _ready() -> void:
 
 	Root.connect("world_origin_shifted", self, "_on_world_origin_shifted")
 
+	_port_v1_to_v2_chunks()
 	_port_to_new_chunk_system()
```
DasCapschen: Shouldn't you re-order them?
If something has to be converted to the new chunk system first, and then ported to the v2 system?

src/Editor/Scripts/Editor.gd
```diff
@@ -95,6 +96,142 @@ func _port_very_old_trackinfo():
...
+		rails[rail.name] = chunk_manager.position_to_chunk(rail.global_transform.origin)
+
+	# Move track objects
+	for track_object in track_objects:
+		if not track_object.attached_rail in rails:
+			Logger.err("Rail %s was not found in %s." % \
+					[track_object.attached_rail, track_object], self)
+			continue
+		var chunk = rails[track_object.attached_rail]
+		if not (chunk in chunks):
```
DasCapschen, line 222: can't you use `if chunk not in chunks` ? 🤔

src/Editor/Scripts/Editor.gd
```diff
@@ -95,6 +96,142 @@ func _port_very_old_trackinfo():
...
+
+		track_object.get_parent().remove_child(track_object)
+		chunks[chunk][0].get_node("TrackObjects").add_child(track_object)
+		track_object.owner = chunks[chunk][0]
+		track_object.update()
+
+	# Save (Resume chunking)
+	world.set_meta("chunk_version", 2)
+	save_world(false)
+	#chunk_manager.resume_chunking()
```
DasCapschen, line 233: why not resume chunking?

DasCapschen (changes requested):
generally looks good. just a few minor comments

src/Data/Scripts/chunk_manager.gd
HaSa1002: Because the Y coordinate has to be 0. So if I'd use snapped, I've the vector with Vector(1, 0, 1). Seemed wasted to me, tbh..

src/Data/Scripts/chunk_manager.gd
HaSa1002: Yeah probably. But with the current implementation the loading distance is one chunk less than the unloading distance which should be fine, I guess

src/Editor/Scripts/Editor.gd
HaSa1002: eh maybe. You did inverse order, too. Hence I followed xD

src/Editor/Scripts/Editor.gd
HaSa1002: Nope

src/Editor/Scripts/Editor.gd
HaSa1002: Because save_world already does that

src/Editor/Scripts/Editor.gd
HaSa1002: I guess the idea was that you only convert them one step at a time?

src/Data/Scripts/chunk_manager.gd
DasCapschen: fair enough

src/Data/Scripts/chunk_manager.gd
DasCapschen: I'd rather we just implement the extra variable. It's basically no work to do it, but we won't have to refactor stuff in the future 😅

src/Editor/Scripts/Editor.gd
DasCapschen: huh... you're right, I did it in reverse ... that was not on purpose 🙈 
we should reorder them to try and run everything in one go... if it breaks, we can do it one step at a time

src/Editor/Scripts/Editor.gd
DasCapschen: okay, my bad. Been using python for too long 🙈

src/Editor/Scripts/Editor.gd
HaSa1002: Well it'll break because you don't save afaik. In all honesty, I'd like to not change that. There are maybe six maps in existence right now. Not worth the trouble.

src/Data/Scripts/chunk_manager.gd
HaSa1002: fair.

HaSa1002: fixed.

src/Editor/Scripts/Editor.gd
HaSa1002: I changed the order but prevented it from executing directly afterwards because as I said, the chunks are not yet saved and because of that the next conversion step will make it a mess

src/Data/Scripts/chunk_manager.gd
```diff
@@ -243,7 +265,8 @@ func resume_chunking():
 
 	set_process(true)
 	loader.set_process(true)
-	loader.load_chunks(get_3x3_chunks(active_chunk))
+	if active_chunk:
+		loader.load_chunks(get_chunks(active_chunk, ProjectSettings["game/gameplay/chunk_unload_distance"]))
```
DasCapschen: still using `chunk_unload_distance` :)

src/Editor/Scripts/Editor.gd
DasCapschen: Chunks should immediately be saved to file upon conversion... 🤔 
Anyways, not really important, like you said there's basically no maps that need it 😄

DasCapschen (dismissed):

src/Data/Scripts/chunk_manager.gd
HaSa1002: done.

DasCapschen (approved):
lgtm

PR #457, opened 2023-01-28
Auto Hide Zs3/Zs3v on Ks-Signals

Lennart1514:
-Hides the Zs3 on Presignals
-Hides the Zs3v on Main Signals


DasCapschen (approved):
Looks good to me. 
Thank you very much for your contribution! ❤️

PR #463, opened 2023-01-29
Fix loader crashing if loading count is wrong

HaSa1002:
I found that bug while I converted the built-in worlds and manually readded resources that were deleted. I didn't change the load_steps at the top and thus it went wrong.


nalquas (approved):
Lgtm!

PR #467, opened 2023-01-29
Add CI-able build version

### LINKED ISSUE #209: macOS Version can't launch
macOS Version doesn't work, .app file is damaged.

grafik

### LINKED ISSUE #371: Automatically set the version in CI
### Summary
Our CI builds lacks a version info. This should be the commit hash

### Godot version hash parsing (in scons)
This is code could be easily ported to GDScript and executed when the build is triggered.
```python
githash = ""
    gitfolder = ".git"

    if os.path.isfile(".git"):
        module_folder = open(".git", "r").readline().strip()
        if module_folder.startswith("gitdir: "):
            gitfolder = module_folder[8:]

    if os.path.isfile(os.path.join(gitfolder, "HEAD")):
        head = open(os.path.join(gitfolder, "HEAD"), "r", encoding="utf8").readline().strip()
        if head.startswith("ref: "):
            head = os.path.join(gitfolder, head[5:])
            if os.path.isfile(head):
                githash = open(head, "r").readline().strip()
        else:
            githash = head
```

HaSa1002:
# Autoversion 
Basically, it asks git for the current ref and tag and calculates a version string as well as some convenience stats like branch being dirty or broken.
It falls back to the version set in the project settings if there's no git.
It also ensures that the project setting changes are as invisible to git as possible by calculating them only when the project is played or exported and resetting them immediatly once it was closed. I hope to prevent `project.godot` change spam...

Fixes #371 

*Edit: Made it an addon now https://github.com/HaSa1002/godot-build-version. I still need to make a logo and upload it to the asset library.*

# CI Updates

As mentioned down below, we have to update our CI pipeline anyway so I took a shot and got rid of the container action. We finally have ad hoc signed Mac builds that people should be able to open with good will. Using native mac builds actually sped up our pipeline considerably because the headless linux builds seem to fail to properly cache *all* imports. Sadly, we can't use the normal editor build, because Godot fails to acquire a non existent X11 interface hence the reason the headless builds exist. So I'd consider this a Godot bug until further notice but I am not patient enough to dig into the issue and file a bug report and fix.

The situation on Windows isn't that much better. I have no idea how to get that natively working. You get like the first line of output and then it hangs or it outputs stuff and Godot returns while still executing and thus the job finished before the export is done. I decided to drop the ball for now (which means no rcedited binaries but that's fine IMHO) and let the MacOS builder export for windows, too. The two builders should be about the same speed, even though the latter one has to export two more builds. 

It's not really and issue and I basically moved the Windows builds to be generated on a Mac from being generated on a dockerised ubuntu ¯\\_(ツ)_/¯

Fixes #209

# CI Future

I left the Windows builds as disabled job for now. I feel like that's fixable but not a top priority right now. I'd like to add automatic uploads to itch.io but we have to decide if we use another itch.io page and in which frequency we gonna update it. I have a copyable action snippet ready. I'd favour updates as they come without scheduling. On the topic of using a new page or the old, however, I've no idea. Both have their benefits and drawbags. If fail to decide, I am gonna toss a coin to your witcher and that's gonna be the decision we gonna stick to.

On this very topic is also release distribution. I feel like we are this close and we should start considering our infrastructure to ease that process. I think about adding a dedicated workflow run on tag or release that gets the latest release of our maps and adds them to the archives. Maybe adding a system that checks commits for keywords and creates a changelog accordingly. Idk that's dreaming, really.

Lastly I'd like to run the potential unit tests on CI, too.

### DIFF
--- a/.github/workflows/export.yml
+++ b/.github/workflows/export.yml
@@ -5,209 +5,256 @@ on:
   pull_request:
     branches: [ master ]
 
+concurrency:
+  group: ci-${{github.ref}}-test
+  cancel-in-progress: true
+
 env:
-  GODOT_VERSION: 3.5
+  GODOT_VERSION: 3.5.1
   EXPORT_NAME: Libre_TrainSim
 
 jobs:
-# TODO: Use native export platforms and add code sign/adhoc notarization
   export-windows:
-    name: Windows Export (Release)
-    runs-on: ubuntu-latest
-    container:
-      image: barichello/godot-ci:3.5
+    if: false # Disabled because @HaSa can't wrap his head around Godot not wanting to properly start and terminate.
+    name: Windows Exports
+    runs-on: windows-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
-      - name: Load Cache
-        uses: actions/cache@v2
+        uses: actions/checkout@v3
+
+      - name: Load Import Folder
+        uses: actions/cache/restore@v3
         with:
           path: src/.import
-          key: ${{github.job}}-${{env.BASE_BRANCH}}-${{github.ref}}-${{github.sha}}
+          enableCrossOsArchive: true
+          key: Import-Windows-${{hashFiles('src/.import')}}
           restore-keys: |
-            ${{github.job}}-${{env.BASE_BRANCH}}-${{github.ref}}-${{github.sha}}
-            ${{github.job}}-${{env.BASE_BRANCH}}-${{github.ref}}
-            ${{github.job}}-${{env.BASE_BRANCH}}
-        continue-on-error: true
-      - name: Setup
+            Import-Windows-${{hashFiles('src/.import')}}
+            Import-Windows-
+            Import-
+      
+      - name: Download Godot
+        run: Invoke-WebRequest -o ../Godot_v${{env.GODOT_VERSION}}-stable_win64.exe.zip  https://github.com/godotengine/godot/releases/download/${{env.GODOT_VERSION}}-stable/Godot_v${{env.GODOT_VERSION}}-stable_win64.exe.zip
+      
+      - name: Extract Godot
+        run: 7z e ../Godot_v${{env.GODOT_VERSION}}-stable_win64.exe.zip -o".."
+      
+      - name: Create Export Template Directory
+        run: mkdir "$env:appdata/godot/templates/"
+      
+      - name: Get Export Templates from Cache
+        id: export-templates-cache
+        uses: actions/cache@v3
+        with:
+          path: ../Godot_v${{env.GODOT_VERSION}}-stable_export_templates.tpz
+          key: Godot_v${{env.GODOT_VERSION}}-stable_export_templates.tpz
+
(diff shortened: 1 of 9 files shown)


HaSa1002:
I fought the last couple of hours against unwanted `override.cfg` issues. It's gonna get a meme xD. I decided to blacklist a few things right away... Let's see. I plan to move the addon into my own repo but developed it here. But y'know there's a little more involved to make and publish such an addon.

Expect a dirty branch version when you download one of those builds. The export_presets are not up to date.

HaSa1002:
Oh lol. That's a first

```
fatal: detected dubious ownership in repository at '/__w/Libre-TrainSim/Libre-TrainSim'
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/Libre-TrainSim/Libre-TrainSim
```

I need to update our pipeline anyway as it will break on May 31st (see the warnings in the actions log). I may as well do that in this pull request, too

HaSa1002:
Note for myself: The download of Godot fails on Windows for some reason.
Note 2: Maybe the headless build is causing the missing cache stuff.

HaSa1002:
Oh only 42 force pushes so far xD
PR is ready to go.
*43

.github/workflows/export.yml
```diff
@@ -5,209 +5,256 @@ on:
...
+          path: src/.import
+          enableCrossOsArchive: true
+          key: Import-Linux-${{hashFiles('src/.import')}}
 
   export-mac:
-    name: Mac Export (Release)
-    runs-on: ubuntu-latest
-    container:
-      image: barichello/godot-ci:3.5
+    # Yes, macOS has the most stable support... Contrary to linux, the caching works, we can do adhoc-notarisation, and it's faster
```
DasCapschen, line 166: Then why not build everything here, even Linux?

DasCapschen (approved):

.github/workflows/export.yml
HaSa1002: Because each step adds like five minutes. I felt like I shouldn't overly increase the total build time even though the usage in total went down.

DasCapschen: but if right now we're running a Linux Build and a Mac/Win Build... then disabling the Linux Build and having a combined build for everything... then total build time wouldn't increase, would it? 🤔

HaSa1002: It would. Right now you've to wait 25 minutes for all checks to pass. Combining all into one would take 35 minutes in to complete. Like the wait time was like 15 minutes but it took six runners to make all the builds

HaSa1002: Sure I could do that again. The setup overhead is like one minute on Mac.

HaSa1002: I think, I am gonna work on another rewrite in a future PR.

DasCapschen: okay.

HaSa1002:
So the issue I had with Windows may be related to powershell quirks as researched in this issue https://github.com/ruby/setup-ruby/issues/13

PR #468, opened 2023-01-29
Fix random resource noise

HaSa1002:
In preparation for a build pipeline that rejects dirty builds (if and only if Godot plays along), we need to ensure there's not random change noise. This also helps contributors to focus on relevant changes.

This should fix dirty branch version labels.


DasCapschen (approved):
lgtm

PR #470, opened 2023-02-05
Prioritize emergency braking over restrictive mode

### LINKED ISSUE #461: PZB tutorial freeing doesn't work
### Summary
You can't free yourself from an emergency break if you fail to break fast enough behind/at the 500Hz magnet in front of the red signal. The game soft locks there. Later on the tutorial tells one that you can disable the emergency break if you press end but that doesn't work there.

### How to reproduce
1. Run the PZB tutorial
2. Follow instructions until the second orange signal
3. Break under 85km/h for the 1000Hz magnet
4. Break under 65km/h but stay over 45 km/h
5. Observe

### Observed behavior
The train performs an emergency break and you can't free yourself from it once it stopped.

### Expected behavior
The trains performs the break and then you can free yourself.

### Version
 - Game Version: 0.9 (#462 )
 - OS: Windows

### Additional information
I think there's also something wrong with the pause mode in PZB safety system. It blinks when the game is paused and does not while the train stands still.

TrojanerHD:
Closes #461

In the `_process` method of the PZB module, the check for restrictive mode runs periodically. This would override the emergency state that was triggered by the PZB leading to the game not realizing that the emergency brake occurred due to PZB.

Short clip showing that the problem in the tutorial was fixed:

[Peek 2023-02-05 01-46.webm](https://user-images.githubusercontent.com/32276754/216795745-c944ff2d-49a7-4d67-93ca-2a7e2c0c0130.webm)

### DIFF
--- a/src/Data/Scripts/PZBModule.gd
+++ b/src/Data/Scripts/PZBModule.gd
@@ -234,6 +234,9 @@ func _on_reverser_changed(state: int) -> void:
 
 
 func restrictive_mode() -> void:
+	if pzb_mode & PZBMode.EMERGENCY:
+		return
+
 	pzb_mode &= ~PZBMode.MASK_MODE   # disable current mode
 	pzb_mode |= PZBMode.RESTRICTIVE  # enable restrictive
 


DasCapschen (approved):
Good catch! Thank you very much! :)

PR #455, opened 2023-01-14
Fix pause behaviour

Jomosoto:
This fixes some bugs that occur when combining different methods to pause the game:

- When the pause menu is open and you close a message or toggle ingame pause, the game is running in pause menu and paused when the pause menu is closed.
- You can toggle ingame pause when the pause menu is open. That allows some functions (like the reverser) when activating ingame pause and unpauses the game when deactivating ingame pause.

PR #471, opened 2023-02-05
Fix pressing escape in tutorial message leading to unexpected behavior

TrojanerHD:
Fixed problem where pressing Esc while a message box is shown would lead to the game opening the pause menu but continuing the game, leading to a reversed state where the game continues in the pause screen and pauses when closing the pause screen.

Clip showing the state before this change:

[Peek 2023-02-05 01-53.webm](https://user-images.githubusercontent.com/32276754/216795998-b0314b4e-3885-47b7-94ae-908ce8339fdc.webm)

Clip showing that the problem has been fixed:

[Peek 2023-02-05 02-01.webm](https://user-images.githubusercontent.com/32276754/216796228-f53ec0ad-dcb2-42c0-9d17-56b4e857e899.webm)

### DIFF
--- a/src/Data/UI/pause.gd
+++ b/src/Data/UI/pause.gd
@@ -11,9 +11,9 @@ var player: LTSPlayer
 
 
 func _unhandled_input(_event) -> void:
-	if Input.is_action_just_pressed("Escape"):
+	if Input.is_action_just_pressed("Escape") and !get_parent().get_node("TextBox").opened:
 		get_tree().paused = !get_tree().paused
-		visible = !visible
+		visible = get_tree().paused
 		if visible:
 			_saved_ingame_pause = Root.ingame_pause
 			Root.ingame_pause = false
--- a/src/Data/UI/text_box.gd
+++ b/src/Data/UI/text_box.gd
@@ -5,6 +5,7 @@ signal closed
 
 
 var _previous_mouse_mode: int
+var opened := false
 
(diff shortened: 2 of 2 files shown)


HaSa1002:
So we have now two attempts to fix the same issue. See https://github.com/Libre-TrainSim/Libre-TrainSim/pull/455...
Blocking the pause menu is not a good solution though. I do like the less invasive approach though so. I guess a middle ground between the two PRs would be great.

TrojanerHD:
Oops, I should've checked for other PRs! I'm not sure what exactly there is to do now though…

HaSa1002:
Well @Jomosoto can voice his opinion, too and then I'll select a PR. Just make sure you two do indeed speak with each other to minimise further duplicated work. As I stated, I like this PRs less invasive approach but the behaviour of the other one is correct. In the end we'll merge one of the two :) No issue there. It happens :D

Jomosoto:
I also tried it this approach in the beginning, but there's the same problem with `ingame_pause`, that can also cause the reversed state. You would have to check for each the pause menu, the message and ingame pause when one of them unpauses with this approach. I don't know what is less complicated in the end.

HaSa1002:
I am happy to merge your PR @Jomosoto once you removed the editor out of the pause logic

Jomosoto:
Sorry, I couldn't find time for this until now. I think I can do it later today

HaSa1002:
Superseded by #455

PR #476, opened 2023-02-06
Rail debugging tools

HaSa1002:
- Rail: Push position tolerance into constant
- Editor: Show connected rails
- Editor: Show start and end position of rails

I implemented the stuff as well as the hack while creating a scenario for Hainfurt. One rail in the Ostkurve didn't wanna connect even though they snapped. I know it's hiding a bug but I really don't want to touch *that* code. (It's probably because of the slope. idk)

### DIFF
--- a/src/Data/Scripts/Rail.gd
+++ b/src/Data/Scripts/Rail.gd
@@ -16,6 +16,7 @@ var track_objects: Array = []
 
 const MAX_LENGTH: float = 1000.0
 const POSITION_TOLERANCE: float = 0.25
+const ROTATION_TOLERANCE: float = deg2rad(1)
 
 export (float) var start_rot: float  # Radians
 export (float) var end_rot: float  # Radians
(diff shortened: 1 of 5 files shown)


src/Data/Scripts/Rail.gd
```diff
@@ -15,6 +15,7 @@ export (bool) var manual_moving: bool = true
 var track_objects: Array = []
 
 const MAX_LENGTH: float = 1000.0
+const POSITION_TOLERANCE: float = 0.25
```
DasCapschen, line 18: Do you want to make a constant for the radial tolerance as well?
As a bonus, we'd get rid of those `deg2rad()` calculations at runtime, since they're put into a constant 😁

src/Editor/Docks/RailBuilder/RailBuilder.gd
```diff
@@ -31,6 +38,10 @@ func update_selected_rail(node: Node) -> void:
 		$S/Settings/Radius/LineEdit.text = String(node.radius)
 		$S/Settings/Angle/LineEdit.text =  String(rad2deg(currentRail.end_rot - currentRail.start_rot))
 		self.set_tendSlopeData(currentRail.get_tendSlopeData())
+		for rail in currentRail.get_connected_rails(true):
+			conntected_rails.add_item(rail.name)
+		for rail in currentRail.get_connected_rails(false):
+			conntected_rails.add_item(rail.name)
```
DasCapschen, line 44: we should do something about this `true/false` business... it's quite confusing...
In general, we should probably do something about the whole `forward` thing on rails...

DasCapschen (dismissed):
More debugging tools are always appreciated 👍🏻 
Haven't tested it, but since you're using it, I trust it works.

Thank you

src/Editor/Docks/RailBuilder/RailBuilder.gd
HaSa1002: `#define FORWARD true` or `constexpr bool FORWARD = true;` 🙃

HaSa1002: But on a serious note, I agree but I am not gonna change it here. Something for the next release. Though using a boolean to indicate the direction is generally a good idea but the API is a mess. Idk how to solve it easily. Maybe we can get rid of most places. idk. We should discuss refactoring places for the next release soon, probably

DasCapschen: Using a boolean is the problem imo. Anything that isn't literally True / False should be expressed as an Enum. Yes, even if the enum only has 2 values.   
imagine `get_connected_rails(Direction.Forward)` vs `get_connected_rails(true)`

but yeah, no need to do it here, I just wanted to note that we should do something about it some time 😅

HaSa1002: Well yeah. I agree the true is not good but I am convinced using an enum for two state things is a can of worms. I'd rather define two constants to prevent people easily adding new entries to the enum. So my mocking above was not completely unserious xD

DasCapschen: What do you mean easily add entries to an enum? It's no different from adding a constant. Enum is basically a constant with a namespace. Instead, usage of the Constants cannot be enfored, nor type checked.
I can just 
`const DRAGON: bool = true; get_connected_rails(DRAGON)`
like... no.

HaSa1002: The point being the enum is internally an integer and not a boolean. And I don't want more than two states in forward.
Sure I can add constants but by the property itself being a boolean I ensure a two state thing.

DasCapschen: Oh that's what you mean. Well... just don't allow anyone to add an enum value? We have reviews. Also why would anyone do it?

HaSa1002: Because it's possible xD Yeah that's true. Anyway topic for a refactor. Let's get 0.9 outta the door and break the whole thing again

src/Data/Scripts/Rail.gd
DasCapschen: I did this :)

DasCapschen (approved):
lgtm 😜

PR #478, opened 2023-02-07
Fix: Restart scenario while message open

Jomosoto:
This fixes a bug where you get stuck on the loading screen when you hit "Restart scenario" while a message is open.

This happens, because only the "pause_menu" pause factor is set to false when restarting and the "message" factor is left true.


HaSa1002 (approved):
lgtm

HaSa1002:
Thanks ❤️

Issue #484, opened 2023-02-25
Train choice is ignored


nalquas:
### Summary

When choosing another train than the default for a track/route, the player's choice is ignored and the default used anyway.

### How to reproduce

1. Click 'Play'
2. Select a map (e.g. 'Tutorials')
3. Select a scenario (e.g. 'Free Driving')
4. Select another train than the default (such as 'JFR1_Grey')
5. Click 'Play'
6. Go to exterior view (press '2') and look at the train

### Observed behavior

The default train has been loaded. For the above example (Tutorials/Free Driving), you'll see the red train.

### Expected behavior

The player's selected train has been loaded. For the above example (Tutorials/Free Driving), this should be the grey train.

### Version

- Game Version: current master (413117976bc611ee584e660cd210eba104f61c88)

DasCapschen:
Can you try a different scenario that is not a Tutorial? 
The tutorial might force the red train.

nalquas:
> Can you try a different scenario that is not a Tutorial? The tutorial might force the red train.

I also tried U2-Nuremberg (where grey is the default, so I selected red) and I still got the default (grey in this case) train.

nalquas:
Oh wow, the player selection is only applied if a route doesn't set a default train. I'll make a fix PR shortly...

PR #485, opened 2023-02-25
Add PassengerCameras to JFR1_Grey

### LINKED ISSUE #465: JFR1-Grey does not support passenger cameras
### Summary
JFR1-Grey doesn't support switching to a passenger camera by pressing keys 3-8.

### How to reproduce
1. Start a game using JFR1-Grey.
2. Try switching camera by pressing keys 3-8.

### Observed behavior
The camera perspective doesn't change.

### Expected behavior
A passenger camera is activated, just like it is on JFR1-Red.

### Version
 - Game Version: current master branch (5eff94e140650724d326f32f4ec4121550353e13)
 - OS: Linux x86_64

### Additional information
The other cameras work fine: 1 for cockpit view, 2 for external camera, 0 for free camera.

nalquas:
Fixes #465. The issue was that the passenger cameras were never added to JFR1_Grey. The solution seems to be to just copy the cameras from JFR1_Red into JFR1_Grey.

The above *should* work, however in testing the camera switching seemed to be unreliable, working in one direction of U2_Nuremberg but not the other (Hauptbahnhof-Röthenbach is problematic). When it didn't work, I ended up in a black screen (GUI still showing though) without audio somewhere across the map. However, this could be due to the game maxing out my GPU's 6GBs of VRAM on that map, making testing very difficult. ~~I also can't select JFR1_Grey on the testing track due to #484.~~

I would appreciate if someone could test it on their machine, or comment if there's an obvious issue with my solution. It could be that I missed some custom script file for one of the train types 🤷‍♂️

### DIFF
--- a/src/Data/Modules/FreeCamera.tscn
+++ b/src/Data/Modules/FreeCamera.tscn
@@ -7,3 +7,4 @@ pause_mode = 2
 far = 1000.0
 script = ExtResource( 1 )
 fixed = false
+handleWorldOriginShifts = true
--- a/src/Data/Scripts/Camera.gd
+++ b/src/Data/Scripts/Camera.gd
@@ -21,6 +21,9 @@ export var fixed: bool = true
 # whether to apply or not acceleration effect on camera
 export var accel: bool = false
 
+# whether to handle world origin shifts - enable if camera has no parent that handles this!
+export var handleWorldOriginShifts: bool = false
+
 # Saves the camera position at the beginning. The Camera Position will be changed, when the train is accelerating, or braking
 onready var cameraZeroTransform: Transform = transform
 
@@ -42,7 +45,8 @@ func _ready() -> void:
 
 
 func _on_world_origin_shifted(delta: Vector3):
-	translation += delta
+	if handleWorldOriginShifts:
+		translation += delta
 
 
 func _input(event) -> void:
--- a/src/Data/Scripts/Player.gd
+++ b/src/Data/Scripts/Player.gd
@@ -76,6 +76,7 @@ var camera_state: int = CameraState.CABIN_VIEW
 var camera_mid_point: Vector3 = Vector3(0,2,0)
 var cameraY: float = 90
 var cameraX: float = 0
+var passengerCameras := []
 
 var mouseSensitivity: float = 10
 
@@ -252,6 +253,14 @@ func ready() -> void:
 		soll_command = 0
 
 	spawnWagons()
+
+	# HACK: Register player passenger cameras. Refactor in future train refactor!
+	passengerCameras.clear()
+	if name == "Player":
+		for wagon in wagonsI:
+			for node in wagon.get_children():
+				if node.is_in_group("PlayerCameras"):
+					passengerCameras.append(node)
 
 	## Prepare Signals:
 	if not ai:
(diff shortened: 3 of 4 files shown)


HaSa1002:
Ok I found your issue. So it assumes the player is spawned first and apparently that's not happening in Nuremberg. 
```gdscript
	var playerCameras: Array = get_tree().get_nodes_in_group("PlayerCameras")
	for i in range(3, 9):
		if Input.is_action_just_pressed("player_camera_" + str(i)) and playerCameras.size() >= i - 2:
			wagonsVisible = true
			camera_state = i
			playerCameras[i -3].current = true
			$Cabin.hide()
			remove_free_camera()
```
(line 680)
the group has 42 nodes in it. I'd say we register them locally instead of using a node group for them.

nalquas:
> the group has 42 nodes in it.

Sounds like the non-player trains all still carry cameras. Wouldn't it make sense to remove them for NPCs? That would also fix the load order problem. Although, I guess the performance impact of non-active cameras is negligible anyway, plus I could imagine deleting nodes from the train could cause issues with possible future custom trains...

> I'd say we register them locally instead of using a node group for them.

How would we do that? If I understand you correctly, each camera would have to actively find the player node (and filter out the NPC trains) and add itself to a camera list?

HaSa1002:
> > the group has 42 nodes in it.
> 
> Sounds like the non-player trains all still carry cameras. Wouldn't it make sense to remove them for NPCs? That would also fix the load order problem. Although, I guess the performance impact of non-active cameras is negligible anyway, plus I could imagine deleting nodes from the train could cause issues with possible future custom trains...

I wouldn't bother deleting those nodes.

> > I'd say we register them locally instead of using a node group for them.
> 
> How would we do that? If I understand you correctly, each camera would have to actively find the player node (and filter out the NPC trains) and add itself to a camera list?

well the train has to be spawned somewhere. Can be a really hacky solution tbh. We should aim for a refactoring of the train stuff in the next release cycle.

nalquas:
> well the train has to be spawned somewhere. Can be a really hacky solution tbh. We should aim for a refactoring of the train stuff in the next release cycle.

Alright, in that case the simplest approach might be to set the playerCameras array when the player is spawned. I'll try it

nalquas:
I just implemented this locally and it seems there's a different issue that I haven't tracked down yet: The cameras genuinely are thousands of units away from the player train on some maps (U2-Nuremberg Hauptbahnhof-Röthenbach, and now also Hainfurt). Might have something to do with chunks? I have no idea honestly.

Edit: Some more investigation with the remote debugger shows the PassengerCameras to be about 4000 units away on the z-axis on U2-Nuremberg relative to their wagons. Something is definitely wrong here...

nalquas:
Hm, moving the camera back into the train via remote debugger show that everything's working correctly, except of cause the initial position of the camera is completely wrong by about 4000 units. I'll investigate further...

HaSa1002:
The wagons were probably not updated at all. If you enter the external camera once, it may work? The wagons are not updated at all iirc when they are not visible. But can also be a missing update.

nalquas:
No, the wagon positions appear to be correct. Only the cameras relative position to its parent seems to be wrong.

I might have found the issue. Look at this snippet of code in `Camera.gd` (lines 38-45):

```gdscript
func _ready() -> void:
	self.set_process_input(true)
	self.set_process(true)
	Root.connect("world_origin_shifted", self, "_on_world_origin_shifted")


func _on_world_origin_shifted(delta: Vector3):
	translation += delta
```

This world origin shift handling looks very strange considering that the camera is positioned *relative to the wagon*.

HaSa1002:
yeah that may be wrong

nalquas:
I fixed the world origin shift handling issue by adding an exported variable `handleWorldOriginShifts`. The only instance where the shift handling turned out to be desirable is `FreeCamera.tscn`, so I enabled it there.

Everything appears to work now. Feel free to test it yourself, but passenger cameras of both red and grey JFR trains work in all maps now.

You can review now :+1:

HaSa1002 (approved):
lgtm

PR #491, opened 2023-03-02
Fix excessive VRAM use of signals and speed limit signs

### LINKED ISSUE #486: High VRAM usage
### Summary

The game currently uses way too much VRAM, maxing out most consumer GPUs on our current maps (U2_Nuremberg / Hainfurt).

### Some ideas

I suspect the game stores each instance of each loaded resource in VRAM. There's a **lot** of redundancy there, and we can probably fix this. It would likely require significant code refactoring/rewriting, though.

Someone will have to do a deep-dive into our codebase and figure out where and why resources are loaded redundantly, and then come up with a more efficient way of doing it.

There's probably a tool for looking at what's using up VRAM in Godot. That could help us narrow down the issue. For example, there's an integrated 'Video RAM' monitor in Godot's debugger, and you can export the data into a text file. I've also tried [Govira](https://gitlab.com/monnef/govira) for this before, which is a VRAM usage analyzer that takes the Godot debugger's exported text file as input.

### Alternatives

None if there are larger / more detailed maps to come...

We will have to address this sooner or later, and the longer we wait, the more difficult it will become to solve.

### More details

My system:

- I'm using a GTX 1060 6GB. U2_Nuremberg just barely fits in there and works fine, but Hainfurt is unplayable.
- I'm on Linux, but have observed the issue on Windows as well.

The issue has been noticed several times already. Some examples:

- https://github.com/Libre-TrainSim/hainfurt/pull/4#issuecomment-1242978308
- https://github.com/Libre-TrainSim/Libre-TrainSim/issues/98
- Probably this one too, since Android phones have little VRAM: https://github.com/Libre-TrainSim/Libre-TrainSim/issues/213#issue-1045799004

nalquas:
Fixes #486.

The cause for the high VRAM usage turned out to be the speed limits texts on the signals and signs. These were previously generated using viewports and then added as a texture for the material of these signs/signals.

As of Godot 3.5, text in 3D space is possible using `Label3D`. I have refactored all existing signals and signs that used viewports to use Label3D. To do it properly, I have also touched the 3D models where necessary and removed the now unnecessary placeholder rectangles (that previously were replaced ingame with the text textures).

- The touched signal types are: Ks, Lf7, Zs3, Zs3_small
- The touched SpeedLimit sign types are: Lf6, Lf6_small, Zs3v, Zs3v_small
- I also touched `Hektometertafel.tscn`

There is basically no visual difference, however the VRAM usage of the game has gone down significantly. I can now play U2-Nuremberg with about 2GB of VRAM used by the game instead of maxing out my GPU with 6GB. That's true even with dynamic lights enabled, by the way. I have also noticed a significant performance increase (higher and more stable FPS) on my machine, probably in part because the game doesn't have to create textures on-the-fly anymore.

I look forward to your feedback!


nalquas:
For reference, here's a VRAM comparison from my machine. The screenshots you see here compare before and after on the U2-Nuremberg map using the Hauptbahnhof-Röthenbach route. The screenshots were taken basically immediately after starting the scenario.

| Before | After |
| --- | --- |
| Screenshot_20230302_141038 | Screenshot_20230302_141123 |
| Screenshot_20230302_141040 | Screenshot_20230302_141121 |

Here are the settings I used:

Screenshot_20230302_140946

HaSa1002 (approved):
lgtm

HaSa1002:
I just wanna have one last opinion from you @DasCapschen

DasCapschen (approved):
Sorry for the late review, I was in Iceland 😇 
Anyways... Thank you for this change! I completely forgot that Label3D is something that was added!  
How did you find out that the Viewports were creating this huge VRAM usage?

HaSa1002:
> How did you find out that the Viewports were creating this huge VRAM usage?

Well first I tried RenderDoc and it gave thousands of buffers and an average texture size of 500x600px. That didn't tell me anything. So I took path 2...

I watched my GPU VRAM usage in the task manager.
I deactivated humans that shove off about one gigabyte so this wasn't the culprit. I disabled mesh instance rendering (which wasn't visible at all iirc).
I deactivated wagons.
And then I disabled signal rendering. Et voilà there I had it. Then I took a look what may went wrong and soon found the viewport and the heavy usage of signals.

DasCapschen:
Alright, thanks so much for finding this! :+1:

PR #492, opened 2023-03-04
Fix build version not having enough data to create version

HaSa1002:
By default most of the git data is ommitted. Fetching the whole history should fix that.


nalquas (approved):
Lgtm!

PR #494, opened 2023-03-06
Fix message text changed when opening settings

### LINKED ISSUE #482: Message text changed when opening settings ingame
### Summary

When opening the settings ingame in the pause menu while a message is open, the text of the message is changed to "Test Test meine Freunde"

### How to reproduce

1. Start any scenario
2. Play until a message is shown
3. Go into the pause menu and open the settings
4. Close the settings
5. See the error

### Version

- Game Version: master (at 4131179)
- OS: Artix Linux x86_64

Jomosoto:
Fixes #482

When opening a message, only the text was set and not the translation_id. When opening the settings, the language is updated and then the message is updated using the translation_id, which still was the default text

### DIFF
--- a/src/Data/UI/text_box.gd
+++ b/src/Data/UI/text_box.gd
@@ -11,7 +11,7 @@ func message(text: String) -> void:
 	_previous_mouse_mode = Input.get_mouse_mode()
 	assert(_previous_mouse_mode != null)
 	Input.set_mouse_mode(Input.MOUSE_MODE_VISIBLE)
-	$MarginContainer/VBoxContainer/Message.text = text
+	$MarginContainer/VBoxContainer/Message.set_text(text)
 	visible = true
 	Root.set_game_pause("message", true)
 	$MarginContainer/VBoxContainer/Ok.grab_focus()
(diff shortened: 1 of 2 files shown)


src/Data/UI/text_box.gd
```diff
@@ -11,7 +11,7 @@ func message(text: String) -> void:
 	_previous_mouse_mode = Input.get_mouse_mode()
 	assert(_previous_mouse_mode != null)
 	Input.set_mouse_mode(Input.MOUSE_MODE_VISIBLE)
-	$MarginContainer/VBoxContainer/Message.text = text
+	$MarginContainer/VBoxContainer/Message.set_text(text)
```
HaSa1002, line 14: FYI it does work with `text`, too. `set_text` is a setter for `text` that we override

HaSa1002 (approved):
lgtm

src/Data/UI/text_box.gd
Jomosoto: It doesn't get called when I use `.../Message.text = text`. I think you can only override [virtual functions](https://docs.godotengine.org/en/stable/tutorials/scripting/overridable_functions.html).

HaSa1002: Interesting. Perhaps because it ignores the script in the typeless context. Because it does work when using types.

PR #497, opened 2023-03-12
remove globalDict for good

### LINKED ISSUE #400: Remove globalDict from World
World contains a variable `var globalDict := {} ## Used, if some nodes need to communicate globally. Modders could use it. Please make sure, that you pick an unique key_name`

It is used in a few places, but we should remove it and replace it with a better solution if necessary.

DasCapschen:
fixes #400


HaSa1002 (approved):
lgtm

PR #498, opened 2023-03-15
Localize input settings

Jomosoto:
Localize input names in the control section of the settings


DasCapschen (approved):
Why do we have an action for "shift" and why is it mappable? Any idea @HaSa1002 or @nalquas ?

otherwise, lgtm :+1:

HaSa1002:
It shouldn't be mappable and was introduced when I added the camera. Afaik it's only used there.

Jomosoto:
We should make a list what actions are shown in settings. There a other actions like Page up/down that don't make sense to modify and also the order often doesn't make sense.

DasCapschen:
@HaSa1002 Why did you make it an action and not just use the modifier? You can ask for shift using `Input.`, right?

Also I agree with @Jomosoto, we should make a list of mappable actions. Would also allow controlling the order :)
~~I'll write a ticket for it.~~ Done: #499

HaSa1002:
> @HaSa1002 Why did you make it an action and not just use the modifier? You can ask for shift using Input., right?

Controller Support probably. I dunno tbh. It was just quick git blame and I didn't dug far

PR #501, opened 2023-03-19
Version Exporter: Ignore MacOS Key Modifications

HaSa1002:
This will hopefully point out what I misconfigured for clean builds to happen...

### DIFF
--- a/.github/workflows/project.godot.ignores
+++ b/.github/workflows/project.godot.ignores
@@ -1,0 +1,49 @@
+diff --git a\/src\/project\.godot b\/src\/project\.godot
+index \S*\.\.\S* \S*
+--- a\/src\/project\.godot
+\+\+\+ b\/src\/project\.godot
+@@ -\d*,\d* \+\d*,\d* @@ SiFa={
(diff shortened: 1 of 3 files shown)


HaSa1002:
Oh wow... It's because of magic... (see https://github.com/godotengine/godot/issues/35355)

HaSa1002:
I am thinking to add an explicit ignore for this change

HaSa1002:
Ok it's now ignored. Not the nicest solution but it has to suffice as workaround until we upgraded to Godot 4.

HaSa1002:
All downloads should now print the version number in white.

DasCapschen (commented):
This is messed up 🙈 

Why do we even need this dirty vs clean thing again?

HaSa1002:
Well while we don't need it technically, rejecting dirty builds is good practice (see game.ci for example) because it helps to catch changes with unwanted side effect.

DasCapschen (approved):
Well alright. It's hacky, but I guess it's okay 🙈

PR #505, opened 2023-03-22
Fix passengers walking through doors when the other side is open

### LINKED ISSUE #500: Passengers walk through closed doors when wrong doors are open
### Summary

It doesn't matter which doors are open, if the doors on any side are open the passengers enter and leave the train. If you open the doors on the wrong side, the people just walk through the closed doors

### How to reproduce

1. Start any scenario with a station
2. Close the doors and open them on the wrong side
3. Watch the passengers walk through the closed doors

### Version

- Game Version: master (at 292bf45e9d1f587795ad442bd3d1c1a0b9a78d8f)
- OS: Artix Linux x86_64

nalquas:
Fixes #500.

Previously, passengers only checked whether any doors (whether left or right) are open to decide whether to walk or not.

This PR adds a `side` variable to `PassengerDoor`. It is assigned in `Wagon.gd`, which already sorted doors into left and right internally anyway. Now, the passengers simply have to check what side their door is on and whether that side of doors is currently open.

Additionally, `Wagon.gd` now uses the `PlatformSide` enum instead of hardcoding these values.

### DIFF
--- a/src/Data/Scripts/Enums/door_side.gd
+++ b/src/Data/Scripts/Enums/door_side.gd
@@ -1,0 +1,13 @@
+class_name DoorSide
+
+enum {
+	UNASSIGNED = 0,
+	LEFT = 1,
+	RIGHT = 2
+}
+
+enum TypeHint {
+	UNASSIGNED = 0,
+	LEFT = 1,
+	RIGHT = 2
+}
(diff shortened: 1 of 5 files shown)


HaSa1002 (approved):
lgtm, but untested

DasCapschen (approved):
Tested. Works. Thanks for the fix 👍🏻

PR #507, opened 2023-03-22
Optimise Pathfinding

HaSa1002:
I implemented a propper A* pathfinding that searches in the direction of the target. I also prevented a few crashes that happen when you add new stations and auto update mode is enabled.

I couldn't edit Hainfurt at all. With the new pathfinding it's almost instant.

### DIFF
--- a/src/Data/Scripts/World.gd
+++ b/src/Data/Scripts/World.gd
@@ -229,64 +229,67 @@ func update_rail_connections() -> void:
 		rail_node.update_connections()
 
 
-# Ensure you called update_rail_connections() before.
-# pathfinding from a start rail to an end rail. returns an array of dicts with rail nodes and direction
-func get_path_from_to(start_rail: Node, forward: bool, destination_rail: Node) -> Array:
-	var visited_rails = {start_rail: {forward = forward, from = null}}
-	visited_rails = _get_path_from_to_helper(start_rail, forward, visited_rails, destination_rail)
-	if visited_rails.size() == 0:
-		return []
-	return _backtrack_path(visited_rails, destination_rail)
+func get_path_from_to(start_node: Rail, end_rail: Rail, forward: bool) -> Array:
+	var reachable := { start_node: [0, null, forward] }
+	var explored := {}
+
+	while not reachable.empty():
+		# Choose some node we know how to reach.
+		var rail := _choose_node(reachable, end_rail)
+
+		# Don't repeat ourselves.
+		explored[rail] = reachable[rail]
+		reachable.erase(rail)
+
+		# If we just got to the goal node, build and return the path.
+		if rail == end_rail:
+			return _build_path(end_rail, explored)
+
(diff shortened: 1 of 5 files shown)


src/Data/Scripts/World.gd
```diff
@@ -229,64 +229,65 @@ func update_rail_connections() -> void:
...
+				reachable[adjacent] = [explored[rail][0] + adjacent.length, rail, fwd]
 
-		if not loop_detected:
-			var visits := visited_rails.duplicate()
-			visits[rail_node] = {forward = forward, from = start_rail}
-			var outcome = _get_path_from_to_helper(rail_node, forward, visits, destination_rail)
-			if outcome.size() > 0:
-				paths.append(outcome)
+			# If this is a new path, or a shorter path than what we have, keep it.
+			if explored[rail][0] + adjacent.length < reachable[adjacent][0]:
```
DasCapschen, line 261: shouldn't this be an `else if` ?
Because if the we saw the node first time, we already set all these values and don't need to check again :)

DasCapschen (dismissed):
untested, but lgtm 👍🏻

src/Data/Scripts/World.gd
HaSa1002: Good point. Added a continue in the previous if

src/Data/Scripts/World.gd
```diff
@@ -229,64 +229,66 @@ func update_rail_connections() -> void:
...
-	var current_rail = destination_rail
 
-	while current_rail != null:
-		route.append({rail = current_rail, forward = visited_rails[current_rail].forward})
-		current_rail = visited_rails[current_rail].from
+func _build_path(rail: Rail, explored: Dictionary) -> Array:
+	var path := []
+	while explored[rail][1]:
+		path.append({"rail": rail, "forward": explored[rail][2]})
+		rail = explored[rail][1]
```
DasCapschen: You are not appending the start rail!

Example:
```gdscript
rail = end_rail
explored = { 
    end_rail=start_rail, 
    start_rail=null 
}

# iteration 1
while explored[rail]: # explored[end_rail] = start_rail ; True
    path.append(rail)
    rail = explored[rail]

# iteration 2
while explored[rail]: # explored[start_rail] = null ; False
    pass # do nothing

# MISSING:
path.append(rail)

path.invert()
return path  # path = [end_rail]
```

You can also **see** this in the scenario editor:
grafik
(green arrow (start) is on a dark blue rail (not part of the path) ; first greenish rail (part of the path) is the one after)

DasCapschen (changes requested):

src/Data/Scripts/World.gd
HaSa1002: true

DasCapschen (approved):
code lgtm

tested, pretty fast, seems to work

DasCapschen:
Thank you ❤️

PR #509, opened 2023-03-27
Fix modding tools not working due to wrong margins

HaSa1002:
Fixes the modding tools because they were not interactable. Their margins was wrong which led to the buttons being outside the bounds.


DasCapschen (approved):
lgtm 👍🏻 

Also added a success popup after mod export since I found that missing when testing.  
Please check if that's ok @HaSa1002, then you can merge

HaSa1002:
yep looking good.

PR #510, opened 2023-03-27
Improve input map for controllers

Jomosoto:
This is the layout (on an Xbox-controller):

Shoulder buttons:
- R2: Accelerate +
- R1: Accelerate -
- L2: Brake +
- L1: Brake -

ABXY buttons:
- A: SiFa
- B: PZB Ack
- Y: PZB Free
- X: PZB Comm

D-Pad:
- Left: Left door
- Right: Right door
- Up: Close doors
- Down: Horn

Sticks:
- Left down: Front light
- Right down: Inside light

I also added this layout to the OpenBVE layout, because I couldn't a default controller layout (Is there any?)

### DIFF
--- a/src/Data/Scripts/PZBModule.gd
+++ b/src/Data/Scripts/PZBModule.gd
@@ -72,7 +72,7 @@ func _on_settings_changed() -> void:
 	else:
 		pzb_mode = PZBMode.DISABLED
 	set_process(is_pzb_enabled)
-	set_process_unhandled_key_input(is_pzb_enabled)
+	set_process_unhandled_input(is_pzb_enabled)
 	emit_signal("pzb_changed", self)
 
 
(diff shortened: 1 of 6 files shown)


HaSa1002:
I think the layout is sensible for a soft launch. But I have to test it to be sure. So please hold on a little while longer.

HaSa1002:
> 
    R2: Accelerate +
    R1: Accelerate -
    L2: Brake +
    L1: Brake -

This feels unnatural. I'd prefer LT to accelerate and RT to deccelerate. Esp. since breaking is not used in the easy mode.

I dunno It's not perfect yet and feel like there's a lot more work for a proper controller support but I think it's fine for 0.9 and proper support is a 0.10 topic.

HaSa1002 (changes requested):
see comment above

Jomosoto:
> This feels unnatural. I'd prefer LT to accelerate and RT to deccelerate

I swapped it. I was habituated to RT to accelerate because of TSW, but now that you say it, LT makes more sense, because the acc stick is on the left side in germany.

> I dunno It's not perfect yet and feel like there's a lot more work for a proper controller support

Yes absolutely. It's just the basics you need to drive the train after you set it up once.

HaSa1002 (dismissed):
lgtm, but please squash into one

HaSa1002 (approved):
Why didn't I get a notification here. Anyway lgtm.

HaSa1002:
Thank you very much ❤️

PR #511, opened 2023-04-03
Fix crash when deleting linked rail of a parallel rail

HaSa1002:
Found while porting docs. Parallel rails stay useless but that's a different story.

In Jeans words:
```
Maybe you used it already in building some switches. And maybe its complete easy for you how they work, but let me explain to you: If you create a parallel rail then this new rail is very special. It is attached to the "old" rail, and gets his all information from the "parent" rail. You just could adjust the parallel distance to it. That makes it very easy for building parallel tracks, and makes Libre TrainSim very powerful, but it also has it's special characteristics:

- If its "parent" rail is deleted, or even the name is changed, then the parallel rail becomes useless, because its "born" to be completely dependent to its parent rail.
- Also the parallel status of this rail can't be changed. 
- *(But in game it is for trains and all other modules a wholesome rail, but rather still depends internally to the parent rail)*
```

### DIFF
--- a/src/Data/Scripts/Rail.gd
+++ b/src/Data/Scripts/Rail.gd
@@ -116,7 +116,7 @@ func update_parallel_rail_settings() -> void:
 	if parallel_rail_name == "":
 		return
 	parallel_rail = get_parent().get_node(parallel_rail_name)
-	if parallel_rail == null:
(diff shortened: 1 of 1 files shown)


nalquas (approved):
Lgtm

PR #522, opened 2023-08-07
Fixes passengers walking through closing doors

### LINKED ISSUE #506: Passengers walk through doors while they are closing

custodian:
While extra condition check does fix the issue, I would rather prefer to implement door state FSM to properly handle door transitions

Something like this: 
https://github.com/Libre-TrainSim/Libre-TrainSim/commit/5fe031b616834c2cc175135dcefc6adc47ad1c08#diff-1e4d945da9b33de5a1d8ca4f34cd2bbc28c9063b1e6147f8b23fc379326bd095

*Edit:
Fixes #506*

### DIFF
--- a/src/Data/Scripts/Person.gd
+++ b/src/Data/Scripts/Person.gd
@@ -104,7 +104,9 @@ func handleWalk(delta: float) -> void:
 
 func is_assigned_door_open() -> bool:
 	assert(transitionToWagon)
-	return (attachedWagon.lastDoorRight and assignedDoor.side == DoorSide.RIGHT) or (attachedWagon.lastDoorLeft and assignedDoor.side == DoorSide.LEFT)
+	return not attachedWagon.lastDoorsClosing and (
+				(attachedWagon.lastDoorRight and assignedDoor.side == DoorSide.RIGHT) or 
+				(attachedWagon.lastDoorLeft and assignedDoor.side == DoorSide.LEFT))
 
 
 func leave_current_wagon()-> void:


HaSa1002 (approved):
Tbh I don't see that much value in adding three new states. I don't really get what their use would be. Like, what's possible with the FSM that's not with the boolean flags. The thing is, these values should be updated by the door animations so that's a consideration, too. Iirc the passengers enter right away or there's a timer (probably something like jTools call_delayed 😬).

Anyway the change itself is good, though not yet tested by me.

@DasCapschen, @nalquas what do you think? Do one of you have the chance to verify the change?

custodian:
States will restrict and simplify all checks in all cases: whenever you need to verify if something should/allow to happen you need to check exactly 1 place, and not think of flag combination (not to mention that you can forget to update one of the clause if you introduce new flag).

Another thing you mentioned is the animation binding. I agree on this one: to my mind doorsClosingTime should be explicitly set to animation duration if for some reason we need to measure the time; and the door states can be bound to animation itself (transitions on signals). Right away we can track single left or right door only, but the path will be open to track/open/close each door individually in each wagon: imagine grain hopper wagons unloading.

Passengers do indeed start to enter door immediately after you start opening it because there is no flag for opening door 😄 

Yet again, while this pull request could still can be merged, I would like to do a FSM and update some passenger path finding as well: i.e. when door is opening they should be allowed to move/gather at door portal but they have to wait for fully opened door before transiting into wagon.

HaSa1002:
> States will restrict and simplify all checks in all cases: whenever you need to verify if something should/allow to happen you need to check exactly 1 place, and not think of flag combination (not to mention that you can forget to update one of the clause if you introduce new flag).

But we do have a binary state right now?

> Another thing you mentioned is the animation binding. I agree on this one: to my mind doorsClosingTime should be explicitly set to animation duration if for some reason we need to measure the time; and the door states can be bound to animation itself (transitions on signals). Right away we can track single left or right door only, but the path will be open to track/open/close each door individually in each wagon: imagine grain hopper wagons unloading.

I don't get that paragraph.

> to my mind doorsClosingTime should be explicitly set to animation duration if for some reason we need to measure the time

What has measuring the time to do with the `doors_closing_time`?

> and the door states can be bound to animation itself (transitions on signals)

yes. to some degree. It's a bidirectional information flow because the player/wagon is gonna start the animation itself. You could use an animation blend tree/state machine. Though the animations can not emit signals but instead call a function.

> Right away we can track single left or right door only

That's fine. If we need autonomous doors they can track their state independent IMHO.

HaSa1002:
> Yet again, while this pull request could still can be merged, I would like to do a FSM and update some passenger path finding as well: i.e. when door is opening they should be allowed to move/gather at door portal but they have to wait for fully opened door before transiting into wagon.

Yeah I mean it looks like you're basically done with that. What about PRing that and I let @DasCapschen and @nalquas give their opinion.

custodian:
> > States will restrict and simplify all checks in all cases: whenever you need to verify if something should/allow to happen you need to check exactly 1 place, and not think of flag combination (not to mention that you can forget to update one of the clause if you introduce new flag).
> But we do have a binary state right now?

Well yes and no. These flags sometimes treated as target state (you will end up there soon, so it basically okay to say that you are already there) and sometimes as terminal state (yes, the doors are closed right now). 
That is the reason why #506 happened.

> > Another thing you mentioned is the animation binding. 
> I don't get that paragraph.

I meant that train model already has animation `open`. And there should no dedicated `doorsClosingTime` setting available for the train, as those are the same. 
`doorsClosingTime` duration should be equal to `open` animation duration (or the maximum from side doors if sim will ever track individual doors).

> What has measuring the time to do with the `doors_closing_time`?

2 things: 
1) how doors are checked right now? `doorsClosingTimer` measures the time to set the door closed.
2) Whenever in the future you will need some more complex calculations, for example: doors are closing and passengers had to estimate if they will be able to finish the transition, or turn around.
There should be

> Though the animations can not emit signals but instead call a function.

Model use `AnimationPlayer` for animation, and it can emit `animation_started`/`_finished`/`_changed` which can be used to track door state transitions.

> That's fine. If we need autonomous doors they can track their state independent IMHO.

And it would be easier to do without arrays of flags to check.

> What about PRing 

I will create another PR with the proposal described above 👍

HaSa1002:
> Well yes and no. These flags sometimes treated as target state (you will end up there soon, so it basically okay to say that you are already there) and sometimes as terminal state (yes, the doors are closed right now).
That is the reason why #506 happened.

Convinced. :+1:

HaSa1002 (approved):
verified in game.
Passengers stop once doors start closing.
Passengers will not try to enter again when doors are reopend. But I think that's a different issue.

src/Data/Scripts/Person.gd
```diff
@@ -104,7 +104,9 @@ func handleWalk(delta: float) -> void:
 
 func is_assigned_door_open() -> bool:
 	assert(transitionToWagon)
-	return (attachedWagon.lastDoorRight and assignedDoor.side == DoorSide.RIGHT) or (attachedWagon.lastDoorLeft and assignedDoor.side == DoorSide.LEFT)
+	return not attachedWagon.lastDoorsClosing and (
+				(attachedWagon.lastDoorRight and assignedDoor.side == DoorSide.RIGHT) or 
+				(attachedWagon.lastDoorLeft and assignedDoor.side == DoorSide.LEFT))
```
HaSa1002, line 109:
```suggestion
	return not attachedWagon.lastDoorsClosing and ( \
			(attachedWagon.lastDoorRight and assignedDoor.side == DoorSide.RIGHT) or \
			(attachedWagon.lastDoorLeft and assignedDoor.side == DoorSide.LEFT))
```
FYI that would be the correct formatting. But since you gonna rewrite that anyway and I approved that now, twice, I will merge as is

HaSa1002:
Thanks ❤️ and congrats to the first merged contribution 🎉

src/Data/Scripts/Person.gd
custodian: Or brackets. Got it. Missed that from the style guide.

custodian:
> Passengers will not try to enter again when doors are reopend. But I think that's a different issue.

Such behavior caused by the fact that as soon as `current_station_table_entry.departure_time` passes, `leave_current_station` is called, `player.current_station_node` becomes `null` and passengers ignore the train.

Raise a separate issue to make correct track of train within station based on `whole_train_in_station` (and similar)?

HaSa1002:
Yeah we can do that

custodian:
> Yeah we can do that

I will be glad to address it after doors ;)

PR #525, opened 2023-08-10
Spawn first rail when chunk_0_0 unloaded

### LINKED ISSUE #438: Creation of rails fails far away from (0,0,0) crashes editor
If the position (0,0,0) is in a unloaded chunk, the creation of new rails crashes.

At the time of the referencing `rail.global_transform.origin` the value is (0,0,0)

grafik

custodian:
Fixes #438 

Steps to reproduce:

1. Create new world
2. Move camera away from chunk_0_0 so it unloads (i.e. to 4 0 with visibility of 2 chunks)
3. Save world
4. Load world. Move further 1 chunk to trigger chunk_0_0 loaded by default to unload
5. Try to place first rail within the map

Core reason is that preloaded rail origin is used and it lead to chunk which is not available.

Proposal is to use current camera origin right away.

### DIFF
--- a/src/Editor/Scripts/Editor.gd
+++ b/src/Editor/Scripts/Editor.gd
@@ -1028,7 +1028,8 @@ func _spawn_poles_for_rail(rail: Node) -> void:
 
 	rail.track_objects.append(track_object)
 
-	var chunk_pos = $World.chunk_manager.position_to_chunk(rail.global_transform.origin)
+	var camera_position = $Camera.global_transform.origin
+	var chunk_pos = $World.chunk_manager.position_to_chunk(camera_position)
 	var chunk_name = $World.chunk_manager.chunk_to_string(chunk_pos)
 	var chunk = $World/Chunks.get_node(chunk_name)
 


HaSa1002:
I think this solution is only fixing a symptom and not the core bug while introducing undeterministic save behaviour (we already have some but I want to get red of that completely because it creates versioning noise and easily breakes your maps)

HaSa1002:
CC @DasCapschen

nalquas:
I'm not sure, but wouldn't using the camera origin mean that track objects could end up in the wrong chunk? Especially if the camera is in one chunk and the rail in another, e.g. due to being close to a chunk's edge?

If I understand correctly, the issue is that the rail origin may not be set (e.g. is 0,0,0). The function `_spawn_poles_for_rail` is called as part of `_spawn_rail`, which instances the rail. However, the rail position is only set after calls of `_spawn_rail`, so may not be set at the time `_spawn_poles_for_rail` is run.

Here's an example of `_spawn_rail` being used (taken from Editor.gd lines 734-741):

```gd
var rail2: Node = _spawn_rail()
rail2.translation = selected_object.end_pos
rail2.start_pos = selected_object.end_pos
rail2.rotation.y = selected_object.end_rot
rail2.start_rot = selected_object.end_rot
rail2.radius = 0
rail2.length = y_length - x_length
rail2.update()
```

Perhaps we should adjust `_spawn_rail` to take a position as a parameter and have it set the translation before `_spawn_poles_for_rail` is called? I'm not sure what else would be affected though.

custodian:
> track objects could end up in the wrong chunk

This is a valid remark, let me dig into this

DasCapschen:
I'm with @nalquas on this one. `_spawn_rail()` should probably specify the rail position right away.  
Also, iirc, we added the chunk of the rail to its metadata at some point. Maybe this should happen in `_spawn_rail()` too. Then we do not have to calculate the chunk for the rail's track objects.   

There's an Edge Case also, which I'm not sure is currently covered:
- if the rail itself is moved, the chunk must be recalculated and all track objects must be moved to the new chunk...
- instead of doing that every move, do it on save ?

HaSa1002:
> instead of doing that every move, do it on save ?

@DasCapschen I thought it's done on save already? Or do I confuse that with the chunk manager and the conversion?

PR #527, opened 2023-08-12
Stop polling if chunk loaded and time not elapsed

custodian:
If chunk has been loaded quickly we still have some loops but `_loader` will be `null`


HaSa1002 (approved):
lgtm. Tested and verified

PR #528, opened 2023-08-13
Door states and Person navigation improvements

### LINKED ISSUE #506: Passengers walk through doors while they are closing

custodian:
Fixes #506 

This is a preview PR as Person Navigation improvements should be verified first.

To improve general design, state tracking and build a foundation for future freight trains where you might need to track door state for each wagon individually.

Using new door states system, persons will be able to:
- Approach opening doors (not yet fully opened) during on-/off-boarding.
- Transition through the doors only when doors are fully opened.
- Idle walk at the station while waiting for the train.
- Get back to the seat and choose another destination if person missed the station.
- Detect crowdy places so they not collide in a single model. 

It might be a good idea to raise a separate enhancement issue to improve the on-/off-boarding priorities.

### DIFF
--- a/src/Data/Scripts/Enums/door_state.gd
+++ b/src/Data/Scripts/Enums/door_state.gd
@@ -1,0 +1,92 @@
+class_name DoorState
+
+# Description
+#
+# CLOSED: 
+#	Door is fully closed.
+#	New Person transition is forbidden.
+# OPENED: 
+#	Door is fully opened.
+#	New Person transition is allowed.
+# OPENING: 
+#	Door is trasiting to OPENED state.
+#	No pathfinding allowed.
+# CLOSING: 
+#	Door is trasiting to CLOSED state. 
+#	New Person transition is forbidden. 
+#	Person already is transition is allowed to continue.
+# HALTED: 
+#	Door is not transitining. 
+#	New Person transition is forbidden. 
+#	Animations/SFX should stop. 
+#	Special state for power cut-off (pantograf).
+
+# Door state transitions:
+#
+#	init 	-> CLOSED
+#	CLOSED  -> OPENING
+#	OPENING -> OPENED | HALTED | CLOSING
+#	CLOSING -> CLOSED | HALTED | OPENING
+#	OPENED  -> CLOSING
+#	HALTED  -> OPENING | CLOSING
+
+enum State {
+	HALTED,
+	CLOSED,
+	OPENED,
+	CLOSING,
+	OPENING
+}
+
+var state: int = State.CLOSED
+
+signal state_changed(state) # int, enum value
+
+func open() -> void:
+	if state == State.OPENED:
+		assert(false, "Doors are already in opened state.")
+		return
+	_set_state(State.OPENING)
+
+
+func is_opened() -> bool:
+	return state == State.OPENED
+
+
+func close() -> void:
+	if state == State.CLOSED:
+		assert(false, "Doors are already in closed state.")
+		return
+	_set_state(State.CLOSING)
+
+
+func is_closed() -> bool:
+	return state == State.CLOSED
+
+
+func is_halted() -> bool:
+	return state == State.HALTED
+
+
+func is_opening() -> bool:
+	return state == State.OPENING
+
+
+func is_closing() -> bool:
+	return state == State.CLOSING
+
+func _init() -> void:
+	state = State.CLOSED
+
+func _set_state(newState: int) -> void:
+	state = newState
+	emit_signal("state_changed", state)
+	
+
+func _on_animation_transition_finished(animation: String) -> void:
+	if animation == "open":
+		match(state):
+			State.CLOSING:
+				_set_state(State.CLOSED)
+			State.OPENING:
+				_set_state(State.OPENED)
--- a/src/Data/Scripts/Player.gd
+++ b/src/Data/Scripts/Player.gd
@@ -49,11 +49,13 @@ var real_arrival_time: int # Time is set if train successfully arrived
 var route_distance_at_station_begin: float = 0 # distance of the station begin on route
 const GOODWILL_DISTANCE: float = 10.0 # distance the player can overdrive a station, or it's train end isn't in the station.
 
+## Doors: 
+# We assume that all doors behave the same way
+# So there is no need to track individual wagons door states
+# Wagons will share doorState state
 export var doors := true # Defines, if this train has doors.
-export var doorsClosingTime: float = 7
-var doorRight: bool = false # If Door is Open, then its true
-var doorLeft: bool = false
-var doorsClosing: bool = false
+var door_left: DoorState = null
+var door_right: DoorState = null
 
 export var brakingSpeed: float = 0.3
 export var brakeReleaseSpeed: float = 0.2
@@ -197,17 +199,6 @@ func ready() -> void:
 		currentSpeedLimit = spawn_point.initial_speed_limit
 	else:
 		currentSpeedLimit = speedLimit
-	if station_table.size() > 0 and station_table[0].stop_type == StopType.BEGINNING:
-		match world.get_signal(station_table[0].station_node_name).platform_side:
-			PlatformSide.NONE:
-				pass
-			PlatformSide.LEFT:
-				open_left_doors()
-			PlatformSide.RIGHT:
-				open_right_doors()
-			PlatformSide.BOTH:
-				open_left_doors()
-				open_right_doors()
 
 	if station_table.size() > 0:
 		current_station_table_entry = station_table[current_station_table_index]
@@ -220,11 +211,6 @@ func ready() -> void:
 		pantograph = true
 		control_type = ControlType.COMBINED
 		reverser = ReverserState.FORWARD
-
-	if not doors:
-		doorLeft = false
-		doorRight = false
-		doorsClosing = false
 
 	if not electric:
 		pantograph = true
@@ -262,6 +248,10 @@ func ready() -> void:
 				if node.is_in_group("PlayerCameras"):
 					passengerCameras.append(node)
 
+	# Open doors at the beginning
+	if station_table.size() > 0 and station_table[0].stop_type == StopType.BEGINNING:
+		auto_open_doors_at_station()
+		
 	## Prepare Signals:
 	if not ai:
 		set_signalWarnLimits()
@@ -346,7 +336,7 @@ func _process(delta: float):
 		check_security()
 
 	if doors:
-		check_doors(delta)
+		check_doors()
 
 	check_signals()
 
@@ -495,7 +485,7 @@ func getCommand(delta: float) -> void:
 		blockedAcceleration = false
 	if command < 0 and not Root.EasyMode and not ai:
 		blockedAcceleration = true
-	if (doorRight or doorLeft):
+	if not are_doors_closed():
 		blockedAcceleration = true
 	if reverser == ReverserState.NEUTRAL:
 		blockedAcceleration = true
@@ -872,7 +862,8 @@ func check_station(delta: float) -> void:
 
 	# If the player is in station and starts driving very early
 	if speed != 0:
-		if is_in_station and not (doorLeft or doorRight):
+		# TODO: confirm that we don't actually need to check the doors as brake is forced when doors are open
+		if is_in_station and are_doors_closed():
 			score -= SCORE_PENALTY_DEPART_EARLY
 			send_message("YOU_DEPARTED_EARLIER")
 			leave_current_station()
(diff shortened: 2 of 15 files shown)


src/Data/Scripts/Enums/door_state.gd
```diff
@@ -0,0 +1,90 @@
...
+	OPENING
+}
+
+var state: int = State.CLOSED
+
+signal state_changed(state) # int, enum value
+
+func open() -> void:
+	if state == State.OPENED:
+		return
```
HaSa1002: Isn't that hiding a bug and warrenting an assert?

src/Data/Scripts/Enums/door_state.gd
```diff
@@ -0,0 +1,90 @@
...
+	_set_state(State.OPENING)
+
+
+func is_opened() -> bool:
+	return state == State.OPENED
+
+
+func close() -> void:
+	if state == State.CLOSED:
+		return
```
HaSa1002: same as above

src/Data/Scripts/Enums/door_state.gd
```diff
@@ -0,0 +1,90 @@
...
+
+enum State {
+	HALTED,
+	CLOSED,
+	OPENED,
+	CLOSING,
+	OPENING
+}
+
+var state: int = State.CLOSED
```
HaSa1002, line 41:
```suggestion
var state := State.CLOSED
```
Use inferred typed when possible

src/Data/Scripts/Player.gd
```diff
@@ -49,11 +49,13 @@ var real_arrival_time: int # Time is set if train successfully arrived
...
+# We assume that all doors behave the same way
+# So there is no need to track individual wagons door states
+# Wagons will share doorState state
 export var doors := true # Defines, if this train has doors.
-export var doorsClosingTime: float = 7
-var doorRight: bool = false # If Door is Open, then its true
-var doorLeft: bool = false
-var doorsClosing: bool = false
+var doorLeft: DoorState = null # State for left side doors
+var doorRight: DoorState = null # State of right side doors
```
```suggestion
var door_left: DoorState = null
var door_right: DoorState = null
```
underscore_case is the case xD I'd remove the comment as it's in the type name

src/Data/Scripts/Player.gd
```diff
@@ -921,19 +908,7 @@ func check_station(delta: float) -> void:
...
-					open_left_doors()
-				PlatformSide.RIGHT:
-					open_right_doors()
-				PlatformSide.BOTH:
-					open_right_doors()
-					open_left_doors()
-			jEssentials.call_delayed(1.2, self, "send_message", [tr("WELCOME_TO") + " " + current_station_table_entry.station_name])
-		else:
-			send_message(tr("WELCOME_TO") + " " + current_station_table_entry.station_name + lateMessage)
+		jEssentials.call_delayed(1.2, self, "send_message", [tr("WELCOME_TO") + " " + current_station_table_entry.station_name + lateMessage])
```
HaSa1002: Please no more call_delayed usages

src/Data/Scripts/Player.gd
```diff
@@ -1036,58 +1009,118 @@ func send_message(string : String, actions := []) -> void:
...
+			call(fnClose)
 		else:
-			open_left_doors()
-	if Input.is_action_just_pressed("doorRight") and not ai:
+			call(fnOpen)
+
+
+func check_doors() -> void:
+	if ai:
+		return
```
HaSa1002, line 1098: AI trains should open their doors, probably

src/Data/Scripts/Wagon.gd
```diff
@@ -12,6 +12,9 @@ var distance_on_rail: float = 0
 var distance_on_route: float = 0
 var speed: float = 0
 
+var doorLeft: DoorState = DoorState.new()
+var doorRight: DoorState = DoorState.new()
```
HaSa1002:
```suggestion
var door_left := DoorState.new()
var door_right := DoorState.new()
```

src/Data/Scripts/Wagon.gd
```diff
@@ -214,14 +197,78 @@ func despawn() -> void:
...
-			if child.translation[2] > 0:
-				child.translation += Vector3(0,0,0.5)
-				child.side = DoorSide.RIGHT
-				rightDoors.append(child)
-			else:
-				child.translation -= Vector3(0,0,0.5)
-				child.side = DoorSide.LEFT
-				leftDoors.append(child)
+			if child.side == DoorSide.UNASSIGNED:
+				# If Door side is not set explicitly, fallback to translation
```
HaSa1002, line 201: I mean... That should be default probably? I don't wanna set these 99% of the time

src/Data/Scripts/Enums/door_state.gd
custodian: If there will be an explicit key bind to open/close doors, and since we have jump to station and force close doors right now.
Assert can be added thought so it will be caller responsibility to check if doors are already in the desired position

src/Data/Scripts/Wagon.gd
custodian: And generally you don't have to, I have assigned only single door on each side within JFR1 to demonstrate this.

The thing is, you might have different wagon designs and/or open/close animations, and rely on a negative vs positive value of a translation of a single point might not be the best idea, so why not allow explicit door allocation?

src/Data/Scripts/Player.gd
custodian: Yes, but `check_doors` only used to check for input, and was verifing that doors has been fully closed

Autopilot does door management, closes before any movement when train departed, and `auto_open_doors_at_station` when arrived.

Persons movement first, and then autopiloting/ai checks before converting draft to pr

src/Data/Scripts/Player.gd
custodian: So, basically, whenever definition is touched, or the vast majority of variable/function usage => convert from camelCase to snake_case to follow the style guide?

HaSa1002: I'd say it depends but in general, yes.

src/Data/Scripts/Enums/door_state.gd
HaSa1002: So it's not assertable? I just rather have have asserts and crashes than silent function returns.

src/Data/Scripts/Player.gd
custodian: I thought that was intentional, as `jTools` and `jEssentials` both have delayed calls (one is processed while paused, the other is not).
I have this `call_delayed` refactoring on my list already.
And here it can be safely omitted.

HaSa1002: It's baggage and don't get me started on that. Because it breaks calling assumptions

custodian: We are on the same page then :)

src/Data/Scripts/Enums/door_state.gd
custodian: But unfortunately it wont work this way with godot3.5: 
` Parse Error: The assigned value doesn't have a set type; the variable type can't be inferred.`

HaSa1002: ah well unlucky xD

src/Data/Scripts/Player.gd
```diff
@@ -1036,58 +1009,118 @@ func send_message(string : String, actions := []) -> void:
...
+		PlatformSide.LEFT:
+			return is_doors_opened(DoorSide.LEFT)
+		PlatformSide.RIGHT:
+			return is_doors_opened(DoorSide.RIGHT)
+		PlatformSide.BOTH:
+			return is_doors_opened()
+	# Platform requires no doors to operate
+	return true
+
+func is_all_doors_closed() -> bool:
```
HaSa1002: Just a comment on the English here. I'd call them `are_doors_*`

src/Data/Scripts/Player.gd
```diff
@@ -1036,58 +1009,118 @@ func send_message(string : String, actions := []) -> void:
...
-	doorsClosingTimer = doorsClosingTime - 0.1
+	_call_wagon_doors("force_close_doors")
 
 
-func check_doors(delta: float) -> void:
-	if Input.is_action_just_pressed("doorClose") and not ai:
-		jAudioManager.play_game_sound("res://Resources/Sounds/click.ogg")
-		close_doors()
-	if Input.is_action_just_pressed("doorLeft") and not ai:
+func toggle_door_side(doorState: DoorState, fnOpen: String, fnClose: String) -> void:
```
```suggestion
func toggle_door_side(door_state: DoorState, fn_open: String, fn_close: String) -> void:
```
Still underscore case

src/Data/Scripts/Wagon.gd
```diff
@@ -214,14 +197,78 @@ func despawn() -> void:
...
+			
+			match child.side:
+				DoorSide.RIGHT:
+					child.translation += Vector3(0,0,0.5)
+					rightDoors.append(child)	
+				DoorSide.LEFT:
+					child.translation -= Vector3(0,0,0.5)
+					leftDoors.append(child)
+				_:
+					assert(true, "Unsupported DoorSide. DoorSide.BOTH is not yet supported")
```
HaSa1002, line 215: I think it'll never be supported xD You can't really be a door on both sides of the train. For freight we probably wanna use a new and different system

src/Data/Scripts/Wagon.gd
```diff
@@ -214,14 +197,78 @@ func despawn() -> void:
...
+
+	# Connect door state to animations
+	var _res = $Doors/DoorLeft.connect("animation_finished", door_left, "_on_animation_transition_finished")
+	_res = $Doors/DoorRight.connect("animation_finished", door_right, "_on_animation_transition_finished")
+
+
+func _animate_side_door(animation: AnimationPlayer, sound: AudioStreamPlayer3D, backwards: bool = false) -> bool:
```
HaSa1002:
```suggestion
func _animate_side_door(animation: AnimationPlayer, sound: AudioStreamPlayer3D, backwards := false) -> bool:
```
FYI that works here, too

src/Data/Scripts/Wagon.gd
```diff
@@ -214,14 +197,78 @@ func despawn() -> void:
...
+	if not animation.is_playing():
+		if backwards:
+			animation.play_backwards("open")
+		else:
+			animation.play("open")
+		# TODO: We should be able to blend left and right door opening sound
```
HaSa1002, line 228: agreed. But that can stay a todo for the time being. The sounds should be played by the door animation IMO

src/Data/Scripts/Wagon.gd
```diff
@@ -214,14 +197,78 @@ func despawn() -> void:
...
+func close_left_doors() -> void:
+	_close_side_doors(door_left, $Doors/DoorLeft)
+
+
+func close_right_doors() -> void:
+	_close_side_doors(door_right, $Doors/DoorRight)
+
+
+func force_close_doors() -> void:
+	# There should be better way to handle this, i.e. queue action?
```
HaSa1002: Not sure I understand

HaSa1002 (commented):
Left a few comments for a final round. Dunno if you wanna tackle them before a merge. Rebasing gonna be painful, I guess

src/Data/Scripts/Wagon.gd
custodian: From my point of few - current doors should work just fine out of the box, but individual wagon doors operation is a must. It will evolve at some point, I hope

Assert here is to ensure that if enum changed we have to review behavior

src/Data/Scripts/Wagon.gd
custodian: Since we have binding between the animations and states, force-set-state concept now is not a best thing to do as internals and visuals might be desynced.

On Train level, `force_close_doors` used in debug mode, when we force jump, and for scenarios specific goals.

We can safely skip everything debug related. And force jump close door part can be safely omitted as we will end up opening doors anyway to pick up passengers and autopilot/enforced brakes will make train close the doors before departure.

Scenarios on the other hand should have no knowledge about subsystems internals, and use only public interfaces. In order to ensure doors are closed - a `close_door` command should issued and signal/check on doors state performed.

I'll add TODO: for a removal of this `force_close_doors` functions

HaSa1002: Ah I see

HaSa1002 (commented):
Code looking good 👍 

Test:
Free Driving: I can't open the doors on arrival in the red train

HaSa1002 (commented):
Test:
Verified in Tutorialmap with Grey and Red. The abovementioned bug was my stupidity. However no passengers enter the grey train

HaSa1002 (approved):
Ok. Can't repro the grey train bug.
Looking good to me ✅ 
Please squash the commits

HaSa1002:
Thanks ❤️

PR #529, opened 2023-08-19
Persons FSM and navigation update

### LINKED ISSUE #490: Passengers seemingly ignore world origin shifts
### Summary

I think passengers ignore any world origin shifts, so they end up somewhere far away from their station on shifts.

During normal play, you probably won't notice this bug as the passengers end up outside of the track. It could be relevant for performance, though.

### How to reproduce

This is probably slightly difficult to reproduce, but in theory:

1. Play a large map (such as Hainfurt)
2. Fly to the other end of the map with the free camera until a world origin shift is triggered (~5000 units?). You could also try to achieve this with the debug camera.
3. Search for passengers near the new world origin

### Observed behavior

There are passengers in the middle of nowhere. They are slowly walking towards their targeted despawn point, aka their station.

### Expected behavior

Passengers stay at their stations and aren't teleported around on world origin shifts.

### Version

- Game Version: current master (594ed462c2e6a3fe3e8f522662feea89f169bbc1)

### Additional information

Screenshot taken on U2-Nuremberg (Hauptbahnhof - Röthenbach).

Screenshot_20230228_201212

### LINKED ISSUE #506: Passengers walk through doors while they are closing

custodian:
Resolves #506
Fixes #490 

This Pull request requires ( built on top of new Door States system PR https://github.com/Libre-TrainSim/Libre-TrainSim/pull/528 )
To simplify review changes related to persons navigation only, I have opened draft pull request to Doors branch https://github.com/custodian/Libre-TrainSim/pull/1/files

This PR also includes a few fixes required:
88f74cd70ce8a0f27d95b6d6d77bac01c3f5fbf3 - Player and Wagon might be out of sync on world shifts and `process()` calls
067814b2052d32e8aeca60cd8dc61448d09f8a35 - Dedicated plugin to draw 2d labels - used for Passengers Debug UI

7cdcdf031f99d33c077eb4775fda1f728128a1be - Not essential, but camera related and fixes issue with camera being drugged in Free Cursor mode.

### DIFF
--- a/src/Data/Scripts/Person.gd
+++ b/src/Data/Scripts/Person.gd
@@ -1,138 +1,281 @@
+class_name Person
 extends Spatial
 
-export (float) var walkingSpeed: float = 1.5
+export (float) var walking_speed: float = 1.5
 
-var attachedStation: Spatial = null
-var attachedWagon: Spatial = null
-var attachedSeat: Spatial = null
-var assignedDoor: Spatial = null
-var destinationIsSeat: bool = false
+var _attached_station: Spatial = null
+var _attached_wagon: Spatial = null
+var _attached_seat: Spatial = null
+var _assigned_door: Spatial = null
 
-var transitionToWagon: bool = false
-var transitionToStation: bool = false
+enum Destination {
+	NONE, # Initial state
+	DOOR_OUTSIDE, # Station -> Door
+	DOOR_INSIDE, # Seat -> Door
+	SEAT, # Door -> Seat
+	SEAT_IDLE, # Seat -> Seat
+	STATION, # Door -> Station
+	STATION_IDLE, # Station -> Station
+}
 
-var status: int = 0
-#0 Walking To Position
-#1: Sitting
+enum Action {
+	IDLE,
+	WALK,
+	SIT
+}
 
-var destinationPos := []
+var _destination: int = Destination.NONE
+var _destination_path := []
+const _destination_tolerance := 0.1
+var _action: int = Action.IDLE
 
-var stopping: bool = false # Used, if for example doors where closed to early
-
-var debug_color := Color(randf(), randf(), randf(), 0.75)
+var _debug_color := Color(randf(), randf(), randf(), 0.75)
 
 func _ready() -> void:
-	walkingSpeed = rand_range(walkingSpeed, walkingSpeed+0.3)
+	walking_speed = rand_range(walking_speed, walking_speed+0.3)
+	_destination = Destination.STATION_IDLE
 
 
 func _process(delta: float) -> void:
-	handleWalk(delta)
+	_handle_walk(delta)
 
 
-var leave_wagon_timer: float = 0
-func handleWalk(delta: float) -> void:
-	# If Doors where closed to early, and the person is at the station..
-	if transitionToWagon and not is_assigned_door_open():
-		if attachedWagon.player.current_station_node != attachedStation:
-			transitionToWagon = false
-			destinationPos.clear()
-		else:
-			stopping = true
-			if $VisualInstance/AnimationPlayer.current_animation != "Standing":
-				$VisualInstance/AnimationPlayer.play("Standing")
-	else:
-		stopping = false
-
-	if destinationPos.size() == 0:
-		if transitionToWagon:
-			attachedStation.deregisterPerson(self)
-			attachedStation = null
-			transitionToWagon = false
-			attachedWagon.registerPerson(self, assignedDoor)
-			assignedDoor = null
-		if transitionToStation and attachedWagon.player.whole_train_in_station \
-				and attachedWagon.is_any_doors_opened():
-			leave_wagon_timer += delta
-			if leave_wagon_timer > 1.8:
-				leave_wagon_timer = 0
-				transitionToStation = false
-				leave_current_wagon()
-		if destinationIsSeat and attachedSeat != null and \
-				translation.distance_to(attachedSeat.translation) < 0.1:
-			destinationIsSeat = false
-			rotation.y = attachedSeat.rotation.y + (0.5 * PI)
-
-			## Animation Sitting
-			$VisualInstance/AnimationPlayer.play("Sitting")
-		elif !$VisualInstance/AnimationPlayer.is_playing() and attachedSeat == null:
-			$VisualInstance/AnimationPlayer.play("Standing")
-
-		return
-
-	if !$VisualInstance/AnimationPlayer.is_playing():
-		$VisualInstance/AnimationPlayer.play("Walking")
+func _on_world_origin_shifted(delta: Vector3) -> void:
+	# Although person is not an World Object,
+	# it is still affected by world origin shift when routing outside of the train
+	if not _is_destination_train_bound():
+		for i in range(_destination_path.size()):
+			_destination_path[i] += delta
 
 
-	if transitionToStation or destinationIsSeat:
-		if translation.distance_to(destinationPos[0]) < 0.1:
-			destinationPos.pop_front()
-			return
-	elif global_transform.origin.distance_to(destinationPos[0]) < 0.1:
-		destinationPos.pop_front()
-		return
-
-	if stopping:
-		return
-
-	var vector_delta := Vector3.ZERO
-
-	if transitionToStation or destinationIsSeat:
-		translation = translation.move_toward(destinationPos[0], delta*walkingSpeed)
-		vector_delta = destinationPos[0] - translation
-	else:
-		global_transform.origin = global_transform.origin \
-				.move_toward(destinationPos[0], delta*walkingSpeed)
-		vector_delta = destinationPos[0] - global_transform.origin
-
-	_debug_draw_path()
-	if vector_delta.z != 0:
-		if vector_delta.z > 0:
-			rotation.y = atan(vector_delta.x/vector_delta.z)
-		else:
-			rotation.y = atan(vector_delta.x / vector_delta.z) + PI
-
-
-func is_assigned_door_open() -> bool:
-	assert(transitionToWagon)
-	match assignedDoor.side:
-		DoorSide.RIGHT:
-			return attachedWagon.door_right.is_opened()
-		DoorSide.LEFT:
-			return attachedWagon.door_left.is_opened()		
+func _is_destination_reached() -> bool:
+	if not _destination_path.empty():
+		var current_pos := translation if _is_destination_train_bound() else global_transform.origin
+		return current_pos.distance_to(_destination_path[0]) < _destination_tolerance
 	return false
 
 
-func leave_current_wagon()-> void:
-	destinationPos.append(assignedDoor.global_transform.origin)
+func _handle_walk(delta: float) -> void:
+
+	_debug_draw_state()
+
+	# Handle next _destination point if required. if we change _destination type, exit.
+	match _destination:
+		Destination.NONE:
+			assert(false, "Some person has no _destination. Spawn failure?")
+			return
+		Destination.DOOR_OUTSIDE:
+			# We supposed to board the train, but it's gone.
+			if _attached_wagon.player.current_station_node != _attached_station:
+				_destination = Destination.STATION_IDLE
+				return
+			# Person is ready to board the train
+			if _destination_path.empty() and is_assigned_door_open():
+				# If wagon is full, restart boarding
+				if not try_board_assigned_wagon():
+					if try_board_train():
+						return
+		Destination.DOOR_INSIDE:
+			# Supposed to exit on this station, but train departed already
+			if _attached_wagon.player.current_station_node != _attached_station:
+				navigate_to_assigned_seat()
+				return
+			# We reached the door
+			if _destination_path.empty():
+				# If opened we can transit, overwise idle
+				if _attached_wagon.player.whole_train_in_station and is_assigned_door_open():
+					if try_leave_current_wagon():
+						return
+		Destination.STATION:
+			# Proceed to next point
+			if _destination_path.empty():
+				_destination = Destination.STATION_IDLE
+				return
+		Destination.SEAT:
+			if _destination_path.empty():
+				_destination = Destination.SEAT_IDLE
+		Destination.SEAT_IDLE:
+			rotation.y = _attached_seat.rotation.y + (0.5 * PI)
+		Destination.STATION_IDLE:
+			# Just walk around and wait for the train
+			if not _attached_station:
+				# Something gone really bad
+				assert(false)
+				return
+			if _destination_path.empty():
+				_destination_path.append(_attached_station.get_random_transform_at_platform().origin)
+
+	if _is_destination_reached():
+		_destination_path.pop_front()
+		return
+
+	if _destination_path.empty():
+		match _destination:
+			Destination.SEAT_IDLE:
+				_action = Action.SIT
+			_:
+				_action = Action.IDLE
+	else:
+		_action = Action.WALK
+
+	# TODO: - check if it is too crowdy around and stop even if supposed to walk
+	# something like get all collisions (if any other person heading same direction then stop)
+	# If there will be problems inside wagons -> limit scope to same destinations only
+	# _action = Action.IDLE or return to keep the animations playing
+	# adjust walking speed accordingly
+
+	_update_action_animation()
+
+	# No need to move
+	if _action != Action.WALK:
+		return
+
+	# Move Person if required
+	var vector_delta := Vector3.ZERO
+	if _is_destination_train_bound():
+		translation = translation.move_toward(_destination_path[0], delta*walking_speed)
+		vector_delta = _destination_path[0] - translation
+	else:
+		global_transform.origin = global_transform.origin \
+				.move_toward(_destination_path[0], delta*walking_speed)
+		vector_delta = _destination_path[0] - global_transform.origin
+
+	# Set rotation torwards destination pos
+	if vector_delta.z != 0:
+		if vector_delta.z > 0:
+			rotation.y = atan(vector_delta.x / vector_delta.z)
+		else:
+			rotation.y = atan(vector_delta.x / vector_delta.z) + PI
+
+	# Adjust rotation based on WorldObject
+	if not _is_destination_train_bound():
+		rotation.y = rotation.y - _attached_station.rotation.y
+
+	_debug_draw_path()
+
+
+func _update_action_animation() -> void:
+	match _action:
+		Action.SIT:
+			_update_visual_animation("Sitting")
+		Action.IDLE:
+			_update_visual_animation("Standing")
+		Action.WALK:
+			_update_visual_animation("Walking")
+
+
+func _update_visual_animation(animation: String) -> void:
+	if $VisualInstance/AnimationPlayer.current_animation != animation:
+		$VisualInstance/AnimationPlayer.play(animation)
+
+
+func is_assigned_door_open() -> bool:
+	assert(_destination == Destination.DOOR_OUTSIDE or _destination == Destination.DOOR_INSIDE)
+	match _assigned_door.side:
+		DoorSide.RIGHT:
+			return _attached_wagon.door_right.is_opened()
+		DoorSide.LEFT:
+			return _attached_wagon.door_left.is_opened()
+	return false
+
+
+# Called by train/wagon when arrived to station
+func arriving_to_station(current_station: Node) -> void:
+	var door_info: Array = _attached_wagon.get_route_from_seat_to_door(_attached_seat)
+	if door_info.empty():
+		assert(false, "Unable to unboard at the station. Side mismatch?")
+		return
+
+	assert(len(door_info) == 2)
+	_attached_station = current_station
+	_assigned_door = door_info[0]
+	_destination_path = door_info[1]
+	_destination = Destination.DOOR_INSIDE
+
+
+func try_board_train() -> bool:
+	var door_info: Array = _attached_station.current_train.get_route_to_free_wagon(global_transform.origin)
+	if door_info.empty():
+		# train do not have free wagons, thats not an issue,
+		# as some persons might still unboarding the train so no assert here
+		return false
+	assert(len(door_info) == 3)
+
+	_attached_wagon = door_info[0]
+	_assigned_door = door_info[1]
+	_destination_path = door_info[2]
+	_destination = Destination.DOOR_OUTSIDE
+
+	return true
+
+
+func try_board_assigned_wagon() -> bool:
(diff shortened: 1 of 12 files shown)


src/Data/Scripts/Wagon.gd
```diff
@@ -58,6 +58,16 @@ func initalize() -> void:
...
 
+	# TODO: FIXME: Make Wagon child of Player/Train.
+	# Player/Train is a WorldObject. RailLogin is a WorldObject.
+	# Wagon is Spatial and not directly bound to Train and relies only on set_transform_on_rail() to set transform.
+	# set_transform_on_rail() is only called during process()
+	#
+	# If there is an world shift happening, all world objects might be shifted, but Wagon will not be notified about it.
+	# This might result in Wagon being in wrong world position, which affect all the calculations relying or global_transform.
+	# This is a temporary fix to make sure that Wagon is always in correct position.
+	Root.connect("world_origin_shifted", self, "_on_world_origin_shifted")
```
HaSa1002, line 67: I assume it messed with the start? Because I don't think it's that much of an issue tbh. They're updated every frame after all

HaSa1002: I assume it messed with the start? Because I don't think it's that much of an issue tbh. They're updated every frame after all

src/addons/basil.debug_labels/debug_label.gd
```diff
@@ -0,0 +1,63 @@
+# Debug labels to output debug information on top of objects
```
HaSa1002: Doc Comments are `##`

src/addons/basil.debug_labels/debug_label.gd
```diff
@@ -0,0 +1,63 @@
...
+var _mount_point: Spatial
+var _mount_offset: Vector3
+
+func _init(mount_point: Spatial, visibility_threshold: float = 50.0, offset: Vector3 = Vector3(0,0,0)):
+	assert(visibility_threshold > 0, "Visibility Threshold should be positive")
+	_visibility_threshold = visibility_threshold
+	_mount_point = mount_point
+	_mount_offset = offset
+	mount_point.add_child(self)
+	owner = mount_point
```
HaSa1002: Why do you set an owner here? I'd avoid that tbh unless you want the label to be saved

src/addons/basil.debug_labels/debug_label.gd
```diff
@@ -0,0 +1,63 @@
...
+#     var test_point:Vector3 = global_transform.origin + Vector3(0, 1.5, 0)
+#     if not cam.is_position_behind(test_point):
+#         if test_point.distance_to(cam.global_transform.origin) < 50.0:
+#             _debug_state_label.visible = true
+#             _position_debug_label(test_point)
+#             _debug_state_label.set_text( "P:%d\n%s\n%s" % [get_instance_id(), Destination.keys()[destination], Action.keys()[action]])
+#         else:
+#             _debug_state_label.visible = false
+
+
```
HaSa1002: Remove in this commit perhaps?

src/Data/Scripts/Person.gd
```diff
@@ -1,148 +1,306 @@
+class_name Person
 extends Spatial
 
-export (float) var walkingSpeed: float = 1.5
+export (float) var _walking_speed: float = 1.5
```
HaSa1002: Well. I'd argue exported variables are public 😅

src/Data/Scripts/Person.gd
```diff
@@ -1,148 +1,306 @@
...
-			## Animation Sitting
-			$VisualInstance/AnimationPlayer.play("Sitting")
-		elif !$VisualInstance/AnimationPlayer.is_playing() and attachedSeat == null:
-			$VisualInstance/AnimationPlayer.play("Standing")
+func _on_world_origin_shifted(delta: Vector3) -> void:
+	# Although person is not an World Object, 
+	# it is still affected by world origin shift when routing outside of the train
+	if not _is_destination_train_bound():
+		for i in range(_destination_path.size()):
+			_destination_path[i] += delta
```
HaSa1002: This callback is starting to get expensive. But okay. Let's roll with it and optimise if needed

src/Data/Scripts/Person.gd
```diff
@@ -1,148 +1,306 @@
...
-		destinationPos.pop_front()
+		Destination.DOOR_OUTSIDE:
+			# We supposed to board the train, but it's gone.
+			if _attached_wagon.player.current_station_node != _attached_station:
+				_destination = Destination.STATION_IDLE
+				return
+			# Person is ready to board the train
+			if _destination_path.empty() and is_assigned_door_open():
+				# Wagon is full, restart boarding
+				if not board_assigned_wagon():
```
HaSa1002: Oh please go to your editor settings and enable strip whitespace

src/Data/Scripts/Person.gd
```diff
@@ -1,148 +1,306 @@
...
+	if _is_destination_train_bound():
+		translation = translation.move_toward(_destination_path[0], delta*_walking_speed)
+		vector_delta = _destination_path[0] - translation
 	else:
 		global_transform.origin = global_transform.origin \
-				.move_toward(destinationPos[0], delta*walkingSpeed)
-		vector_delta = destinationPos[0] - global_transform.origin
-
-	_debug_draw_path()
+				.move_toward(_destination_path[0], delta*_walking_speed)
```
HaSa1002: Please use a space around ops https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html

src/Data/Scripts/Person.gd
```diff
@@ -1,148 +1,306 @@
...
+		return false
+	assert(len(seat_info) == 2)
+	_attached_station.deregister_person(self)
+	_attached_seat = seat_info[0]
+	_destination_path = seat_info[1]
+	_destination = Destination.SEAT
+	return true
+
+# So far we do not expect any fails when leaving the wagon
+func leave_current_wagon() -> bool:
```
HaSa1002: I can't imagine a case when we would

src/Data/Scripts/Player.gd
```diff
@@ -48,6 +48,7 @@ var is_in_station := false # true if the train speed = 0, the train is fully in
 var real_arrival_time: int # Time is set if train successfully arrived
 var route_distance_at_station_begin: float = 0 # distance of the station begin on route
 const GOODWILL_DISTANCE: float = 10.0 # distance the player can overdrive a station, or it's train end isn't in the station.
+var leave_current_station_announced: bool = false # to prevent depature re-announcement till train is refactored with a state
```
HaSa1002, line 51: What a state? I thought stateless is modern 🙃

src/Data/Scripts/Player.gd
```diff
@@ -1107,10 +1113,28 @@ func check_doors() -> void:
...
+			nearestDoorIndex = index
+	return nearestDoorIndex
+
+
 func auto_open_doors_at_station() -> void:
 	if not current_station_table_entry:
 		return
-	var station_node = world.get_signal(current_station_table_entry.station_node_name)
+	var station_node: Node = world.get_signal(current_station_table_entry.station_node_name)
+	assert(station_node != null)
```
HaSa1002: You maybe wanna check using is_instance_valid

src/Data/Scripts/Wagon.gd
```diff
@@ -18,8 +18,8 @@ var door_right := DoorState.new()
 var leftDoors := []
 var rightDoors := []
 
-var seats := [] # In here the Seats Refernces are safed
-var seatsOccupancy := [] # In here the Persons are safed, they are currently sitting on the seats. Index equal to index of seats
+var seats_free := []
+var attachedPersons := {} # Person -> occupied seat reference
```
HaSa1002: underscore case

src/Data/Scripts/Wagon.gd
```diff
@@ -352,63 +353,75 @@ func registerPassengerPathNodes() -> void:
...
+	if route.empty():
+		Logger.err("Some doors are not reachable from every door! Check your Path configuration", self)
+		assert(false)
+		return []
+
+	# Update position of door. (The Persons should stick inside the train while waiting ;)
+	if closest_door.side == DoorSide.LEFT:
+		route[route.size()-1].z += 1.3
+	else:
+		route[route.size()-1].z -= 1.3
```
HaSa1002: Does that break with different trains? Like, that is setting some hard facts how wide the trains are, right?

src/Data/Scripts/Wagon.gd
```diff
@@ -456,3 +469,24 @@ func updateSwitchOnNextChange(): ## Exact function also in player.gd. But these
...
+			_debug_passenger_state_label = null
+		return
+	var total_seats := seats_free.size() + attachedPersons.size()
+	if total_seats == 0:
+		return
+	if not _debug_passenger_state_label:
+		_debug_passenger_state_label = DebugLabel.new(self, 50, Vector3(0, 5, 0))
+
+	if _debug_passenger_state_label.is_visible():
+		_debug_passenger_state_label.set_text( "W:%d\nPassenger: %d / %d" % [get_instance_id(), attachedPersons.size(), total_seats])
```
HaSa1002: One empty line pls

src/Data/Scripts/World.gd
```diff
@@ -68,19 +69,30 @@ func _ready() -> void:
...
+	
+	player.update_waiting_persons_on_next_station()
+
+
+func get_new_person_instance() -> Person:
+	randomize()
+	var person: Person = _person_template.instance()
+	var person_visual: PackedScene = _person_visual_instances[int(rand_range(0, _person_visual_instances.size()))]
+	person.add_child(person_visual.instance())
+	person.owner = self
```
HaSa1002: Not needed as they shouldn't be saved

HaSa1002 (changes requested):
untested eye only review. Dunno if helpful

src/addons/basil.debug_labels/debug_label.gd
custodian: This has been already removed, not sure why this popped up, let me check on this one

src/Data/Scripts/Person.gd
custodian: My bad, yes

src/addons/basil.debug_labels/debug_label.gd
custodian: No intention to save the label, just to make sure destruction occurs whenever owner destroyed, since that's a plugin

src/Data/Scripts/Wagon.gd
custodian: That's the way how it was before those states, proper way would be introducing those shifts within `PassengerDoor`/or train.
Left it as is for now, might be marked as TODO: 
I haven't finished my first train yet, so I don't quite understand what are the pros and cons in terms of support for this one

src/Data/Scripts/Wagon.gd
custodian: Wagons are updated, yes, but if you spawn at the station, and people decide to board, they will receive an invalid route from the station to the door, as rail, station and wagon already spawned and placed, world has been shifted, and that shift moved player, station and persons but not wagons, so first destination point in the path is far far away from the station (initial world shift delta).

We can of course make a sanity check for the destination distance at Person class (i.e. distance/destination should be within station bounds), but this looks like an workaround rather a proper system functioning.

HaSa1002: Hmm. I thought the drive(0) in initialize sets the transform. That's why I am asking

src/addons/basil.debug_labels/debug_label.gd
HaSa1002: Destruction is not by owner but by tree. So when it is in the tree and the parent is deleted so are also the children. Owner have nothing to with that afaik

src/Data/Scripts/World.gd
custodian: Didn't know it might affect `PackedScene`

src/addons/basil.debug_labels/debug_label.gd
HaSa1002: It's removed in the next commit

src/Data/Scripts/Wagon.gd
HaSa1002: Is fine. Was a general remark

src/Data/Scripts/World.gd
HaSa1002: Well only if they are saved

src/addons/basil.debug_labels/debug_label.gd
custodian: I'll amend commits to be clean

src/Data/Scripts/Station.gd
```diff
@@ -34,48 +38,50 @@ func _ready():
...
+	if personSystem:
+		personsNode = Spatial.new()
+		add_child(personsNode)
+		personsNode.owner = self
+
 
 func _process(_delta: float) -> void:
-	handlePersons()
+	handle_persons()
+	_debug_passanger_state()
```
TrojanerHD:
```suggestion
	_debug_passenger_state()
```
Small typo ^^

src/Data/Scripts/Person.gd
```diff
@@ -1,145 +1,306 @@
...
+	if _attached_wagon and _attached_wagon.is_person_registered(self):
+		_attached_wagon.deregister_person(self)
 	queue_free()
 
 
-func clear_destinations() -> void:
-	destinationPos.clear()
+func _is_destination_train_bound() -> bool:
+	match _destination:
+		Destination.SEAT, Destination.DOOR_INSIDE:
```
TrojanerHD: Did you consider adding `SEAT_IDLE` here? Although you don't need it in your case, I imagine if this function might get re-used in other contexts it could lead to confusion if that's not included here

src/Data/Scripts/Person.gd
```diff
@@ -1,145 +1,306 @@
...
+		return false
+	assert(len(seat_info) == 2)
+	_attached_station.deregister_person(self)
+	_attached_seat = seat_info[0]
+	_destination_path = seat_info[1]
+	_destination = Destination.SEAT
+	return true
+
+# So far we do not expect any fails when leaving the wagon
+func leave_current_wagon() -> bool:
```
TrojanerHD: Maybe rename to `can_leave_current_wagon`? It confused me a bit when reading the code that this function does not actually set `_destination` e. g.

src/Data/Scripts/Player.gd
```diff
@@ -48,12 +48,19 @@ var is_in_station := false # true if the train speed = 0, the train is fully in
...
 export var doors := true # Defines, if this train has doors.
-export var doorsClosingTime: float = 7
-var doorRight: bool = false # If Door is Open, then its true
-var doorLeft: bool = false
-var doorsClosing: bool = false
+var door_left: DoorState = null
+var door_right: DoorState = null
+# Doors cache to be used within current station
+var _station_doors: Array = []
+var _station_doors_wagons: Array = []
```
TrojanerHD:
```suggestion
var _station_doors := []
var _station_doors_wagons := []
```

src/Data/Scripts/Person.gd
```diff
@@ -1,145 +1,306 @@
...
-	if transitionToStation or destinationIsSeat:
-		if translation.distance_to(destinationPos[0]) < 0.1:
-			destinationPos.pop_front()
+func _is_destination_reached() -> bool:
+	if not _destination_path.empty():
+		if _is_destination_train_bound():
+			if translation.distance_to(_destination_path[0]) < _destination_tolerance:
+				return true
+		else:
+			if global_transform.origin.distance_to(_destination_path[0]) < _destination_tolerance:
```
TrojanerHD: Really nitpicky and I'm only remarking this because you did it elsewhere that way too:
```suggestion
		elif global_transform.origin.distance_to(_destination_path[0]) < _destination_tolerance:
```

src/Data/Scripts/Station.gd
```diff
@@ -107,38 +113,37 @@ func getRandomTransformAtPlatform() -> Transform:
...
+
+# return destination position for the person
+func register_person(personNode: Spatial) -> Array:
 	attachedPersons.append(personNode)
 	personNode.get_parent().remove_child(personNode)
 	personsNode.add_child(personNode)
-	personNode.owner = world
-	personNode.destinationPos.append(getRandomTransformAtPlatform().origin)
+	personNode.owner = self
+	return [get_random_transform_at_platform().origin]
```
TrojanerHD, line 152: As of now, this function doesn't need to return an array but can instead return the Vector3 directly

src/Data/Scripts/Wagon.gd
```diff
@@ -214,46 +212,111 @@ func despawn() -> void:
...
-func getRandomFreeSeatIndex() -> int:
-	if attachedPersons.size()+1 > seats.size():
-		return -1
-	while (true):
-		var randIndex: int = int(rand_range(0, seats.size()))
-		if seatsOccupancy[randIndex] == null:
-			return randIndex
-	return -1
+
+func get_random_free_seat_index() -> Spatial:
```
TrojanerHD:
```suggestion
func get_random_free_seat() -> Spatial:
```
This actually returns the seat itself, not the index

TrojanerHD (commented):
Didn't extensively test the changes but I have got some remarks mostly about naming schemes

src/Data/Scripts/Person.gd
custodian: may be `try_leave_current_wagon`? and same for `board_assigned_wagon` -> `try_board_assigned_wagon`?

src/Data/Scripts/Station.gd
custodian: While we do not have multiple points right now, path might be more complex in the future.
I would insist on having same return type for all navigation related functions, i.e. if you receive destination, you know it's a path.

src/Data/Scripts/Wagon.gd
custodian: yep, a left over after cleanup

src/Data/Scripts/Person.gd
custodian: We can move it one level up to the station, where we will have 1 connection for all affected station passengers (as persons on the trains are not affected anyway.

HaSa1002: Yeah the world origin shift is a prime candidate for optimisation. I mean, deferring the execution of these things over multiple frames is probably enough and fine.

src/Data/Scripts/Wagon.gd
custodian: At the time of `initialize` and first call to `drive(0)` and `set_transform_on_rail` world shift is set to loaded from map/scenario value, and later it is shifted once again to camera position (to route start point) and that's where wagon desyncs till first `process` call

We can also make this connection `CONNECT_ONESHOT` since it is required only once after spawn (at least for now till we get AI trains and peek function, which might highlight more issues)

src/Data/Scripts/Player.gd
custodian: I would prefer an assert here, as if we have something within `station_table` but it does not have proper signal meta info - something gone completely wrong after `get_calculated_station_points` (scenario start) and before arrival to the station.

But I guess we can actually use `current_station_node` right away. Not sure why I choose to go with `get_signal` in the first place

src/Data/Scripts/Person.gd
custodian: Yep, it should be added here!

Technically, `SEAT_IDLE` and `STATION_IDLE` are not required as separate states for proper navigation and states handling. 

I decided to explicitly add those to highlight that Person might have some other stuff to do while traveling within the wagon (i.e. animated chat/yawn animations from time to time, or even toiler trip), or at the station (i.e. not just walk around but take a sit at the bench, checkout some vending machines/etc.

src/Data/Scripts/Wagon.gd
HaSa1002: Yeah, I'd favor that solution, with a TODO maybe

src/Data/Scripts/Player.gd
HaSa1002: Oh wait. I mean inside the assert. The null check will fail on freed objects that we still point to...
```suggestion
	assert(is_instance_valid(station_node))
```

src/Data/Scripts/World.gd
HaSa1002:
```suggestion
```
Causes owner invalid error

HaSa1002 (changes requested):
Test:
Verified with grey and red train in Tutorial. Working fine except noted comments.
I always get a you departed early yelling message even departing on time and way too late

src/Data/Scripts/Person.gd
custodian: I explicitly use `else: if` to high expressions leading to the statement, so it it will be not miss-interpreted as a precondition if you ever need to refactor or add more branches, so the similar statements are on the same level of indentation. Mostly because project does not have any unit tests yet.

Imagine the following:
```
if A:
  if B:
    do_first()
else:
  if C:
    do_second()
do_overwise()
```

It might be easy to make a mistake during refactor:
```
if A:
  if B:
    do_first()
elif C:
  do_second()
do_overwise()
```
to result in erroneous code:
```
if A and B:
    do_first()
elif C:
  do_second()
do_overwise()
```

To highlight this fact using `elif` and a single indentation level it would be like
```
if A and B:
  do_first()
elif not A and C:
  do_second()
do_overwise()
```
And this is tends to be less readable to assess all branches.

custodian: What do you think of this one?
```
func _is_destination_reached() -> bool:
	if not _destination_path.empty():
		var current_pos = translation if _is_destination_train_bound() else global_transform.origin
		return current_pos.distance_to(_destination_path[0]) < _destination_tolerance
	return false
```

custodian:
I will squash PR Code review before merge, `departure time check` also can be squashed. 
`prevent persons unboarding at stations without platforms` is another bug found with custom map, can be fixed here or as different PR

src/Data/Scripts/Person.gd
HaSa1002: I think makes sense here

HaSa1002:
Retested in Tutorial with red and grey. It doesn't leave the station anymore (I departed a bit late because I was waiting for all passengers to board)
grafik

Also pressing `O` to close the doors in the external view doesn't close the doors. You have to press at least twice to trigger the animation. (Not reproducible with the grey train)
grafik

HaSa1002 (approved):
verified and retested with red ✅ 

All working fine. lgtm. Please make the history merge ready :>

HaSa1002:
Thanks ❤️

PR #533, opened 2023-08-20
Unselect previously selected list item to allow reselection

custodian:
Editor `RailLogicMenu` does not allow reselection.
While we can tick `Allow Reselect` property on `RailLogicMenu`, it is visually better to remove the selection


HaSa1002 (approved):
lgtm and tested ✅

HaSa1002:
Thanks ❤️

PR #535, opened 2023-08-24
Fix Scenario Editor route renaming

### LINKED ISSUE #514: Scenario Editor saving is broken
### Summary
The scenario editor doesn't save the correct or some corrupt state.

### How to reproduce

1. Open or create a scenario (tested with Hainfurt scenarios)
2. Duplicate and rename routes
3. Make some changes
4. Save
5. Restart LTS
6. Open the scenario

### Observed behavior
The duplicated scenario in its original form reappears. When removing and then saving again, an empty string scenario is saved.

### Expected behavior

Duplicating should work, remvoing should work, renaming should work and none of them should reintroduce the og data.

### Version

- Game Version: v0.9.beta1.299.custom.aa9e4b6
- OS: Windows

### Additional information
Likely a cache and resource duplication/ownership issue.

**Edit:** Further investigation revealed the issue to be worsend to a crash if the resource was edited within Godot.

Jomosoto:
Fixes #514 

The cause of the problem is, that after renaming the route is readded to the list while switching to the route under the new name.

Is there any reason why a copy of the route is created when switching routes, so you always work on a copy of the route, that is only applied when you switch routes?
If not, this would be an easy solution

### DIFF
--- a/src/Editor/Scripts/ScenarioConfiguration.gd
+++ b/src/Editor/Scripts/ScenarioConfiguration.gd
@@ -41,8 +41,6 @@ func _input(_event):
 
 
 func save():
-	if is_instance_valid(loaded_route):
-		routes[current_route] = loaded_route.duplicate(true)
 	scenario_editor.scenario_info.routes = routes.duplicate(true)
 	scenario_editor.scenario_info.rail_logic_settings = rail_logic_settings.duplicate(true)
 	scenario_editor.scenario_info.save_scenario()  # write to disk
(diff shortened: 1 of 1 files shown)


HaSa1002:
I don't remember any from the top of my head. 
Verification needed:
1. Creating new route
2. Deleting a route 
3. Adding stops
4. Removing stops
5. Save and load
6. Rename of a route
7. Duplicate and the edit of a route

Jomosoto:
I verified everything. After another change everything works

HaSa1002:
Nice. Self review. That's something new xD I'd still like a second pair of eyes to test this. My comment was more of a general guidance what I'd test. Be assured. If nobody checks until I finished work later today, I'll review and merge :)

HaSa1002 (approved):
Verified above steps. Working as expected :white_check_mark:

HaSa1002:
Thanks ❤️

PR #537, opened 2023-08-26
Iterate through the directory only if its accessible

custodian:
`get_files_in_directory` can cause stack overflow If there is inaccessible directory within the track directory

### DIFF
--- a/src/Editor/Scripts/export_track.gd
+++ b/src/Editor/Scripts/export_track.gd
@@ -37,7 +37,9 @@ static func export_editor_track(track_name: String, export_path: String) -> Stri
 static func get_files_in_directory(path: String) -> Array:
 	var files = []
 	var dir = Directory.new()
-	dir.open(path)
+	if dir.open(path) != OK:
+		return []
+
 	dir.list_dir_begin(true, true)
 	var file_name = dir.get_next()
 	while file_name != "":


src/Editor/Scripts/export_track.gd
```diff
@@ -37,13 +37,13 @@ static func export_editor_track(track_name: String, export_path: String) -> Stri
...
-		else:
-			files.append(path.plus_file(file_name))
+	if dir.open(path) == OK:
+		dir.list_dir_begin(true, true)
+		var file_name = dir.get_next()
+		while file_name != "":
+			if dir.current_is_dir():
+				files.append_array(get_files_in_directory(path.plus_file(file_name)))
+			else:
+				files.append(path.plus_file(file_name))
```
HaSa1002:
```suggestion
	var err := dir.open(path)
	if err != OK:
		return []

	dir.list_dir_begin(true, true)
	var file_name = dir.get_next()
	while file_name != "":
		if dir.current_is_dir():
			files.append_array(get_files_in_directory(path.plus_file(file_name)))
		else:
			files.append(path.plus_file(file_name))
```

I'd prefer an early return and left-bound code

HaSa1002 (changes requested):
Change is good, just a style rule we have

HaSa1002 (approved):
lgtm

HaSa1002:
Thanks :heart:

PR #540, opened 2023-09-17
Remove non-editor tracks from the scenario editor track selection menu

### LINKED ISSUE #456: Crash: "Test track" and "Export Track" in scenario editor
Platform: Linux x86_64

When you click on "Test track" or "Export Track" in the scenario editor pause menu the game get's terminated by a SIGSEGV signal.

When clicking these buttons inside the godot editor the game freezes and uses more and more memory.

Jomosoto:
Fixes #456

Only show editor tracks in the scenario editor selection menu, because other tracks (built-in tracks and mods) can't be modified.

Alternative: Propose the player to create a copy of the map in the editor folder

### DIFF
--- a/src/Editor/Scripts/scenario_editor_selection.gd
+++ b/src/Editor/Scripts/scenario_editor_selection.gd
@@ -23,12 +23,9 @@ func _unhandled_input(event: InputEvent) -> void:
 
 func update_track_list():
 	$TrackList/ItemList.clear()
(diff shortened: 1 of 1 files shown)


HaSa1002 (approved):
I think this is valid. To reenable that we have to store the scenarios in a different location but I think that should be post 0.9 material

HaSa1002:
Thanks ❤️

PR #541, opened 2023-09-18
Fix create new track

### LINKED ISSUE #530: Assertion fails upon opening newly created map in track editor
### Summary

When creating a new track and trying to open it in the track editor, an assertion fails.

### How to reproduce

1. Create a new track ("Create" → "Track Editor" → enter some name → "New")
2. Select the track and press "Open"
3. Observe

### Observed behavior

As loading reaches 100%, the game halts due to a failed assertion with the following output:

Screenshot_20230819_231449

Restarting Libre-TrainSim and opening the track again fixes the map, but shows a message "World was converted to v2 chunks. Please reload the world".

### Expected behavior

The track loads fine.

### Version

- Game Version: Current master (commit 3952171a8222b2f1f2a48b16adf946769ac1ae35)

### Additional information

The mentioned line 927 of Editor.gd reads:

```gd
assert($World/Chunks.get_child_count() == 0)
```

Jomosoto:
Fix #530 

- Add the `chunk_version` metadata (with the value 2) to the world pattern, so the Editor doesn't try to port it
- Fix the failing assert when porting a track, that occurs because after porting a chunk is only added to the  `_loaded_chunks` of the Chunk Loader  when there are duplicates for the chunk coordinates (as a result of the porting)

### DIFF
--- a/src/Editor/Scripts/Editor.gd
+++ b/src/Editor/Scripts/Editor.gd
@@ -157,6 +157,8 @@ func _port_v1_to_v2_chunks() -> void:
 	# Merge possible duplicates
 	for chunk_position in chunks:
 		if chunks[chunk_position].size() == 1:
+			chunks[chunk_position][0].name = ChunkManager.chunk_to_string(chunk_position)
+			chunk_manager.loader._loaded_chunks.append(chunks[chunk_position][0].name)
 			continue
 
 		var best_fit: int = 0
--- a/src/Data/Modules/World-Pattern.tscn
+++ b/src/Data/Modules/World-Pattern.tscn
@@ -8,6 +8,9 @@
 
 [node name="World" type="Spatial"]
 script = ExtResource( 1 )
+__meta__ = {
+"chunk_version": 2
+}
 
 [node name="DirectionalLight" parent="." instance=ExtResource( 2 )]
 


src/Editor/Scripts/Editor.gd
```diff
@@ -192,7 +193,6 @@ func _port_v1_to_v2_chunks() -> void:
 			local_chunk.get_parent().remove_child(local_chunk)
 			local_chunk.free()
 
-		best.name = ChunkManager.chunk_to_string(chunk_position)
```
HaSa1002: That should probably stay 'cause the node may be named chunk0_0@1 or something because of multiple chunks

HaSa1002: But I am not sure. Did you test porting Hainfurt or Nuremburg?

Jomosoto: Oh no, I didn't think of that. Yes, I should have tested it :/

Jomosoto: I tried to test it on Hainfurt and Nurenberg, but I ran into another error problem: `load(path_to_Hainfurt/Nuremberg.tscn)` returns `Null`, even though the path exists (verified with `File::file_exists`). Do you know this issue from the original port?

DasCapschen (approved):
LGTM 👍🏻

src/Editor/Scripts/Editor.gd
HaSa1002: hmm. Then the og file is corrupted for some reason. No worries though. I don't think anyone besides a very select circle is gonna port tracks anyway. Let's just keep aware of the issue

HaSa1002:
Thank you very much ❤️

HaSa1002 (commented):
I approved ofc 😂

PR #547, opened 2023-10-24
fix broken default overhead line

DasCapschen:
When creating a new map, the first rail by default spawns with an overhead line, but without poles!  
This is confusing, since the UI toggle for overhead lines spawnes poles...  

As a fix, I have disabled overhead lines by default, and left it as a setting for the user to enable. This way, poles will always be there when an overhead line is requested, and that without spawning them immediately when a world is created...
This only affects new worlds! If a world was previously created with a floating overhead line, it will remain like that! No overhead lines or poles will disappear from existing maps. 

Also fixes a crash when disabling overhead lines. The RailBuilder (UI) was not updated to the new chunk system I guess 🙈


src/Editor/Scripts/Editor.gd
```diff
@@ -1039,6 +1039,19 @@ func _spawn_poles_for_rail(rail: Node) -> void:
...
+	rail.update_overhead_line(null)
+	
+	var chunk_pos = $World.chunk_manager.position_to_chunk(rail.global_transform.origin)
+	var chunk_name = $World.chunk_manager.chunk_to_string(chunk_pos)
+	var to = $World/Chunks.get_node(chunk_name).get_node("TrackObjects")
+	
+	var node_name = rail.name + " Poles"
+	
+	if to.has_node(node_name):
+		to.get_node(node_name).queue_free()
```
HaSa1002: I like how our code base relies so heavily on predefined node names that we construct in code /s

HaSa1002 (commented):
untested pure code review.
looking good. Someone has to give it a quick test though (not me tonight)

HaSa1002 (approved):
works fine, after the rebase I just did

HaSa1002:
Thanks ❤️

PR #549, opened 2023-12-02
Add render stats to debug window

HaSa1002:
Adds render stats below the debug FPS counter. Can be enabled and disabled using ProjectSettings and an in-game toggle.

I had my reasons...
grafik
grafik

### DIFF
--- a/src/Data/UI/debug_ui.gd
+++ b/src/Data/UI/debug_ui.gd
@@ -4,33 +4,59 @@ extends CanvasLayer
 var debug_camera: DebugCamera
 
 
-onready var fps_label := $DebugContainer/FPSContainer/FPSLabel as Label
+onready var fps_label := $Stats/Infos/FPSLabel as Label
 onready var camera_target_list := $DebugContainer/DebugCameraTargets as ColoredItemList
+onready var stats := $Stats as Control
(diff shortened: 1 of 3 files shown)


DasCapschen (approved):
Looks fine to me. Hope with that we can figure out why it runs so poorly 🙈

HaSa1002:
I am quite impressed it runs at all actually xD I mean 100 million verts a quite a feat

DasCapschen:
Not too many draw calls though... Multimeshes should be the reason for that.
Also a 3080 can do 30 trillion floating point ops a second, 100 million verts should be fine :D

HaSa1002:
I probably have to compare that to a high quality demo scene perhaps. Something with proper PBR mats. I guess the high amount of surface changes contributes? I dunno. Doesn't really look fishy apart the vert count

DasCapschen:
Many changes of things per frame, that seems weird to me, tbh

HaSa1002:
yeah and 1,2k mats is quite a bit. Especially considering we only have like 15 materials and this is an editor running

DasCapschen:
Perhaps we're duplicating multimesh materials somewhere? Or do some materials have "local to scene" checked or something?   
We don't even have that many materials in the project! 😅

PR #550, opened 2023-12-02
Some gizmo improvements

HaSa1002:
I wanted to finish the last station of Hainfurt...
First, when I positioned a building and saved, it was gone and not to be found. Turns out, we move buildings between chunks but don't retain their global position.

Then I wanted to move a building and it was 1000 metres elsewhere. Also rotating and then moving something close to tracks wasn't fun. That's where the local transform mode comes from. And the position input code was such mess that I need to clean it up. Sorry (not sorry) :D

- Gizmo: Fix local to global offset when moving
- Editor: Add local transform mode
- Saving: Ensure buildings stay at their position
- Fix whitespace

PS: Hainfurt wasn't developed today. 🙈

### DIFF
--- a/src/Editor/Editor.tscn
+++ b/src/Editor/Editor.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=31 format=2]
+[gd_scene load_steps=33 format=2]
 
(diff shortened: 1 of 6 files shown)


DasCapschen (approved):
lgtm

only code review, not tested. But since you made it for your own pain, I trust it works 😅

PR #554, opened 2024-03-04
Fix safety systems error with ai trains

Jomosoto:
On ai trains, the Node of type LTSPlayer is not called "Player" and the safety systems get the player node by name.

This fix disables the Saftey Systems if the player node is null, as they shouldn't be enabled in ai trains anyways. This isn't a clean solution, but it's uncomplicated.

Other solutions would be
- to export the player variable (you would have to connect it manually, but the PZB signals also have to be connected manually at the moment)
- or to have it set by the player maybe as part of the SafetySystems class.

The last one is my favourite, as every safety system will have to have access to the player, but I'm not sure and I'd love to hear you opinion


HaSa1002 (approved):
That's a fair change. I agree, we should add these systems on request to the train and simply not add them in that case. So yes, option three

